PR #44, opened 2014-10-07
Adds docopt to the installed python modules

tombettany:
Related to https://github.com/KanoComputing/kano-settings/pull/186

PR #51, opened 2015-02-03
Install docopt for python by default

pazdera:
I would like to use [docopt](http://broken.build/2015/01/19/docopt/) in the kano tracker script - it's a really nice way of building a command line parser.

Also I think @tombettany made a PR to some project to migrate to this that's been hanging there for a few months.

cc @alex5imon


tombettany:
Same as https://github.com/KanoComputing/kano-updater/pull/44. Docopt is pretty awesome - I used it https://github.com/KanoComputing/kano-settings/pull/186 and it makes things a lot simpler but it was never merged

tombettany:
Also, from the Readme on https://github.com/docopt/docopt, they advise using `pip install docopt==0.6.1` to install - i.e. specify the version

alex5imon:
I'll close this and @tombettany  will merge #44 instead

PR #52, opened 2015-02-04
Downgrading the improved FBTurbo driver to the previous version

skarbat:


alex5imon:
Would this remove `/usr/share/X11/xorg.conf.d/999-fbturbo-improved.conf` for all users?

skarbat:
Tested on the PI followed by a reboot. It certainly removes this file, and the previous one is installed: ```/usr/share/X11/xorg.conf.d/99-fbturbo.conf```

pazdera:
I'd maybe put it into PreUpdate rather than post. Technically, it doesn't matter though.

skarbat:
Makes sense @pazdera  . Done.

pazdera:
Thank you, Albert :-).

Issue #50, opened 2015-01-28
image resources not translatable


hzulla:
The updater displays text stored as graphics in its image resources ([examples](https://github.com/KanoComputing/kano-updater/tree/master/images/header)).

To make these translatable, there are two possibilities.

- compose translated images during build using imagemagick scripts
- compose same design using Gtk widgets

What is the preferred way to handle this? I'd go the widget route (saves storage space, too), but the designer might be unhappy. ;-)

hzulla:
What's the name of the font used here?

![header-1](https://cloud.githubusercontent.com/assets/1705654/5961896/f9b3e4dc-a7df-11e4-8cd6-855b7ef8a9f9.png)

It doesn't seem to be Arista or Bariol from the kano-fonts package.

tombettany:
@hzulla I believe that this is Bariol but, regardless, it probably should be even if it isn't (Bariol is the preferred font).

@TommySaehl Am I correct?

hzulla:
Arista:

![arista](https://cloud.githubusercontent.com/assets/1705654/5999845/9612fec8-aad6-11e4-9b25-c0e51b0b2e04.png)

Bariol:

![bariol](https://cloud.githubusercontent.com/assets/1705654/5999846/995485e8-aad6-11e4-8775-ef52b059554b.png)

Compare the "g" at the end.

TommySaehl:
Hm.. Strange..

This is Bariol, but the 'g' is messed up; 
![image](https://cloud.githubusercontent.com/assets/1634790/5999895/324b77d4-aad7-11e4-9810-29aa4398a1b7.png)

Where did you grab that from @hzulla ? 

And yes @tombettany – Bariol, always and forever.

tombettany:
@hzulla Good point. But I'm fairly certain that it should be Bariol.

TommySaehl:
Found it – was a setting in Photoshop that must've been active when doing the Updater assets.

hzulla:
@TommySaehl  - it's from [header-1.png](https://github.com/KanoComputing/kano-updater/tree/master/images/header).

TommySaehl:
@hzulla @tombettany It is Bariol, and the supposedly messed up 'g' isnt't messed up.

hzulla:
(Oh boy, working with Gtk explains why people switch to Qt.)

[...]

PR #53, opened 2015-02-05
replace image resources with Gtk widgets

hzulla:
Hi,

`kano-updater-gui` displays a fake GUI by using text stored as bitmaps in [images](https://raw.githubusercontent.com/KanoComputing/kano-updater/89e2c36a986e053a480e8ced44c35c43f46e8377/images/header/header-5.png), which makes the application [not translatable](https://github.com/KanoComputing/kano-updater/issues/50).

This patch replaces the images with Gtk widgets. It's a close, but obviously not an exact replica of the original layout:

stage5

As requested by @TommySaehl, the labels are using the Bariol font now. It [is not possible to define a line-height](http://stackoverflow.com/questions/25139645/gtk3-label-line-spacing) for Gtk.Label, but imho the layout still looks ok. (I tried using a Gtk.TextView instead, but that resulted in messier code.)

A nice side-effect is that it's now easy to change the poster text copy.

Please note that `current_number` is now 0-5 instead of 1-6 inside `kano-updater-gui`, which made the code easier to read.

What's not to like? The code contains a fair number of hard-coded layout definitions. No idea if these can be moved to the CSS instead.


hzulla:
Note that the psa-label ("DO NOT DISCONNECT") has a different colour compared to the (!) image. This is because the (!) is actually bright white ffffff with transparency, while the label is using CSS lightgray d3d3d3. Needs to be fixed.

tombettany:
Wow @hzulla great work! I'll try this out now.

tombettany:
@hzulla This is AWESOME! Great work - I love the slide transitions.

@carolineclark Any objections to merging this?

@TommySaehl Is this close enough to the original design?

carolineclark:
If it works, then no objections :)

tombettany:
@FMog I'll rebuild the package now. Can you test when this is done?

tombettany:
@FMog Package is built and available in the devel suite whenever you get a chance to update and test.

hzulla:
Please proofread. Typos are possible. :-)

TommySaehl:
Ah @hzulla , looks great man! Its close to shippable, just some tweaks;
The above image is the updater with my changes, and below I point out what changes I've made (haven't change font size though)

image

FMog:
So just tested this, it looks really good. Only one thing I noticed was a slight mistake in the description on step 1:

screenshot 2015-02-05 14 58 07

As you can see it states "Sent it to us at hello@kano.me" when it should be "Send it to us"

Probably not even related to these changes, but it does need changing :)

hzulla:
@TommySaehl ok, should I provide a patch with different pixel values or do you guys just want to sit down and twiddle with them on your own until you're happy?

Here's what you need to change:

* [font sizes and colours in the CSS](https://github.com/KanoComputing/kano-updater/blob/2b078987b20cb5f335ac3b9f9650e2b06d0eb1be/images/CSS/updater.css)
* [top margin of the headline](https://github.com/KanoComputing/kano-updater/blob/2b078987b20cb5f335ac3b9f9650e2b06d0eb1be/bin/kano-updater-gui#L153)
* [bottom margin of the psa label container](https://github.com/KanoComputing/kano-updater/blob/2b078987b20cb5f335ac3b9f9650e2b06d0eb1be/bin/kano-updater-gui#L284)
* [size of the stage label](https://github.com/KanoComputing/kano-updater/blob/2b078987b20cb5f335ac3b9f9650e2b06d0eb1be/bin/kano-updater-gui#L331)
  (the stage label is centered within its box, so there is no "top margin" defined for it)

To make sure that the Gtk layout engine isn't confused, please note that these height values need to add up for the stage box:

[stage box height](https://github.com/KanoComputing/kano-updater/blob/2b078987b20cb5f335ac3b9f9650e2b06d0eb1be/bin/kano-updater-gui#L309) = [stage label height](https://github.com/KanoComputing/kano-updater/blob/2b078987b20cb5f335ac3b9f9650e2b06d0eb1be/bin/kano-updater-gui#L331) + [progress box height](https://github.com/KanoComputing/kano-updater/blob/2b078987b20cb5f335ac3b9f9650e2b06d0eb1be/bin/kano-updater-gui#L340)

The window is currently layouted for 450 px height:

450px height = [header box height](https://github.com/KanoComputing/kano-updater/blob/2b078987b20cb5f335ac3b9f9650e2b06d0eb1be/bin/kano-updater-gui#L133) + [stage box height](https://github.com/KanoComputing/kano-updater/blob/2b078987b20cb5f335ac3b9f9650e2b06d0eb1be/bin/kano-updater-gui#L309)

TommySaehl:
@hzulla Sorry, but don't really know what you mean by "provide a patch with different pixel values".
Would be great if you'd able to implement those tweaks.

Thanks!

hzulla:
I meant that modifying the layout now means changing those pixel values I referenced to in the previous comment.

My RPi is back on Tuesday. Will report back then with a patch.

alex5imon:
@hzulla don't worry, we will do this. Thanks a lot for your work here, it has been extremely useful

hzulla:
OK, thx. :-)

PR #55, opened 2015-02-11
Migrating to the new tracker

pazdera:
Migrated to the new tracker and added some new tracking points. Neil should now be able to tell when people update, from which version and whether the update finished properly.

A few formatting changes on the way.

Related to:
https://github.com/KanoComputing/peldins/issues/1358
https://github.com/KanoComputing/kano-profile/pull/201

cc @alex5imon @skarbat


skarbat:
looks good to me @pazdera Radek !

PR #58, opened 2015-02-13
I18n gettext hooks for kano-updater-gui

hzulla:
See https://github.com/KanoComputing/kano-settings/pull/215 for details.
This patch follows the same recipe.

updater


hzulla:
outdated, replaced by #63.

PR #59, opened 2015-02-17
Squeeze in a 4th stage - Optimising your Kano

tombettany:
KanoComputing/peldins#1582
To provide a more responsive feel, split the package installation stage
between downloading and installing the packages.

@pazdera @skarbat Does this look like it might break anything?


skarbat:
Doesn't look like it's going to hurt @tombettany

tombettany:
@skarbat Great, I'll merge then :-)

PR #61, opened 2015-02-18
Fix lxpanel plugin icon

tombettany:
Moving of the icon files for the GUI resulted in the lxpanel plugin
icons getting moved and installed to a different location. Group these
images with the lxpanel plugin code and install them to the original
location.


tombettany:
@pazdera Does this seem okay?

PR #62, opened 2015-02-19
move fixed layout width/height values into constants

hzulla:
makes it easier to change height/width of the main layout elements and computes PROGRESS_HEIGHT from the height values it depends on (instead of using a fixed value).


tombettany:
@hzulla This is much nicer. I'm happy to merge as soon as we've double checked by testing but It doesn't look like there'll be any problems.

tombettany:
Just tested and it works nicely. Merging now. Good work @hzulla :+1:

PR #63, opened 2015-02-19
I18n gettext hooks for kano-updater-gui, kano-updater and check-for-updates (2nd try)

hzulla:
This replaces #58 and adds i18n to the new modularized kano-updater-gui and its added fourth stage.

updater-gui

See https://github.com/KanoComputing/kano-settings/pull/215 for details.
This patch follows the same recipe.


po/de.po
```diff
@@ -0,0 +1,250 @@
+# German translations for PACKAGE package
+# German messages for PACKAGE.
+# Copyright (C) 2015 THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the PACKAGE package.
```
tombettany: @hzulla Do we need to fill out the actual information here? Also, you say this file is auto-generated; what is the process to update this if we need to add more text?

lxpanel-plugin/kano_updater.c
```diff
@@ -8,7 +8,8 @@
 
 #include <gtk/gtk.h>
 #include <gdk/gdk.h>
-#include <glib/gi18n.h>
+#define GETTEXT_PACKAGE "kano-updater"
+#include <glib/gi18n-lib.h>
```
tombettany: @hzulla Sorry, I'm not too familiar with gettext, do we need to compile this file with all the languages or is this just providing a link to the languages that we've defined?

kano_updater_gui/header.py
```diff
@@ -104,15 +104,15 @@ def __init__(self, number):
             margin_bottom = 25
         )
 
-        if number < 5:
+        if number < NUMBER_OF_STAGES:
```
tombettany: @hzulla Well spotted here

kano_updater_gui/header.py
```diff
@@ -104,15 +104,15 @@ def __init__(self, number):
...
             psa_bang = Gtk.Image()
             psa_bang.set_from_file("{}/bang.png".format(IMAGE_PATH))
             psa_grid.add(psa_bang)
 
         psa_label = Gtk.Label(
-            (u"Do not disconnect your Kano"
-            if number < 5 else
```
tombettany: And here.

debian/install
```diff
@@ -10,3 +10,5 @@ lxpanel-plugin/*.png usr/share/kano-updater/images/
 
 icon/updater.app usr/share/applications/
 icon/kano-updater.png usr/share/icons/Kano/66x66/apps/
+
+locale/* /usr/share/locale
```
tombettany: Sorry, I'm not an expert - is this the standard directory for languages to be installed?

bin/kano-updater-gui
```diff
@@ -150,7 +156,7 @@ class MainWindow(Gtk.Window):
         self.set_stages(current_number)
         self.start_spinner()
 
-    # number is the current working state (1-6) minus one (0-5)!
+    # number is the current working state (1-7) minus one (0-6)!
```
tombettany: Another good spot. I was a bit lazy it would seem...

debian/install
hzulla: Yes. cd po && make will create the right directory structure which includes the package name.

lxpanel-plugin/kano_updater.c
hzulla: You just compile it. The program will read the language file it needs from /usr/locale at runtime. If it doesn't find any for its locale, it doesn't complain and just shows the original strings. In other words, you can distribute this version without language files and it will just work as it does now.

po/de.po
hzulla: No, people seem to ignore these doc comments, anyway. messages.pot is auto-generated from the sources, <lang>.po is updated with new strings from messages.pot with msgmerge. The process is explained in TRANSLATION.md.

debian/install
hzulla: Do a find /usr/locale and you will immediately understand the structure of these .mo files.

tombettany: Ah, I see. Great stuff.

lxpanel-plugin/kano_updater.c
tombettany: Ah, Great! Sounds perfect.

po/de.po
tombettany: Sorry, I missed the merge bit.

po/Makefile
```diff
@@ -0,0 +1,33 @@
...
+endef
+
+messages:
+	$(run-xgettext)
+
+messages.pot:
+	$(run-xgettext)
+
+%.po: messages.pot
+	msgmerge -N -U $*.po messages.pot
```
tombettany: @hzulla I've tried using our servers to build this package but for some reason it hangs on this `msgmerge` line. I'm a little confused though: if we are just building the package, not doing any new translation, should the `.po` files be generated again at all? Or is this for the case where the code has been updated but the translations haven't?

po/Makefile
```diff
@@ -0,0 +1,33 @@
...
+	touch $*.po
```
tombettany: Also @hzulla, what is the purpose of this command? If the `.po` files don't exist then the `msgmerge` command fails anyway and if they do exist then touching them isn't going to do anything.

A little confused.

po/Makefile
tombettany: After a chat with @skarbat, it seems that we encountered this same `msgmerge` problem last year when we were trying to build `openbox`. It is apparently an unsupported QEmu emulation call and the solution before was to just comment out the `msgmerge` line.

@skarbat suggested that this problem may be fixed since then, it was in the `qemu-user-static` package. There is potential that updating this *may* fix the problem but it might disrupt the build environment for our OS build. @pazdera any thoughts?

tombettany: I found the bug report [here](https://bugs.launchpad.net/qemu/+bug/668799) and it claims to have been fixed in June 2013 so you would have thought that we would already have the fix???

po/Makefile
hzulla: msgmerge (in my Ubuntu box) will modify the existing .po file during a merge and then not change its modification timestamp. So that the .po timestamp never changes even if its content does.

po/Makefile
hzulla: I will change that Makefile. The msgmerge is redundant and not really required for build, calling it during build is only a side-effect of the rules that are useful during development. Only msgfmt is needed for packaging.

hzulla:
@tombettany - the new commit contains a super-condensed intro to the gettext basics. Sorry about the msgmerge crash. Will [this](http://forums.gentoo.org/viewtopic-p-7200436.html) help?

Other than that: To build, one only needs to run `msgfmt` to convert .po files to .mo.

But my Makefile comes with rules for development, where the .po files depend on messages.pot. And since that one isn't part of the source ball because it can be generated from the sources, the Makefile does just that. Now messages.pot is newer than the .po file, so it will be "updated" with `msgmerge`, which doesn't change the content of the .po file. See?

Again, I will provide a better Makefile that doesn't do the needless steps. On Monday.

po/Makefile
tombettany: Okay, fair enough.

po/Makefile
tombettany: Yeah, a Makefile that doesn't do `msgmerge` should solve the problem. It is one of those annoying bugs in our build server setup, sorry for the trouble.

tombettany:
I like the gettext basics readme - I certainly had many questions (I'm sorry for that) and if I had this resource then things would have been much clearer. Good for getting up to speed with the development.

I'll have a look into the `msgmerge` tip that you found but I'm not sure how successful I'll be.

I see how the Makefile that you've done would be useful for development. If you want to keep your dev rules then I'm happy to just have a separate rule for `build` which just does the `msgfmt`.

hzulla:
@tombettany ok, the new Makefile will now always create *all* .po files during `make`, but it won't call `msgmerge` anymore. Should work for you.

tombettany:
@hzulla Perfect! It now behaves nicely on the build server. I also successfully installed the package and ran it in both English and German.

A note on the merge conflict: We added [a new rule for going from v1.3.3 to v1.3.4](https://github.com/KanoComputing/kano-updater/blob/997de9b781a5d95263767c322e5b051c0a8adad8/kano_updater/scenarios.py#L272-L276) but you changed [the v1.3.2 to v1.3.3 rule](https://github.com/hzulla/kano-updater/blob/43d492728fde4888e59e1a8fb301bdc03f2b8ef2/kano_updater/scenarios.py#L264-L265). This should be simple to fix, just ensure that the line says

```python
    def beta_132_to_beta_133(self):
        run_cmd_log('LC_ALL=C kano-apps install --no-gui terminal-quest')

    def beta_133_to_beta_134(self):
        pass
```
when you resolve.

One slight thought is that when running the non-English version there is a Gtk warning (only a warning mind you, it runs fine):

```
Gtk-WARNING **: Locale is not supported by C library.
        Using the fallback 'C' locale.
```

This is just because by default we do not add the multi-language support (it can be added interactively by using `sudo dpkg-reconfigure locales` and there is probably a non-interactive version which is as of yet undetermined). At some point we should add this locale generation into our build process but it doesn't need to be done now as it is just a warning and shouldn't be done in this package anyway.

Please also note that unless we set `LC_ALL` to be a different language, everything will behave as it already did so we don't need to fully integrate the language changes until we are ready systemwide.

@alex5imon @pazdera @carolineclark @skarbat @Ealdwulf @convolu I think that this is almost ready to merge (after resolving the conflict). Any final thoughts or concerns on the way the translations are handled here? No matter how small the concerns are, it is best to voice them now while we figure out the perfect recipe.

hzulla:
Nice. Thanks for testing, @tombettany!

Please note that I also replaced ""button_release_event" with "clicked" in one commit, see above. That changes the behaviour of the buttons slightly, but imho improves the GUI. A long-term goal would be to make the Kano UI more keyboard-friendly so that you aren't required to use the mouse for everything.

> At some point we should add this locale generation into our build process

A possible way to do this is to just add en_EN.us to the default image. Only upon first boot, the system should ask the user for the default language, then generate the locale(s) for it and continue with the usual welcome process for the new user, using that language.

That would also make it possible to preset a few default choices (e.g. German keyboard layout and Berlin timezone for de_DE).

Later, a user-specific locale setting could be added to `kano-settings`.

hzulla:
A different possibility is offering language-specific images for download. More work for your release team, but that would leave the initial "huh, what's going on here, what's up with the white rabbit?" sense-of-wonder experience untouched for new users if they don't have to suffer through an initial "please let us know your language" multiple choice menu.

hzulla:
@tombettany - if it's not too much hassle, please resolve the merge conflict on your side if you merge it. If you decide to merge some far later time in the future, I will happily provide a new pull request if the source changes much until then.

pazdera:
Hi @hzulla,

It looks amazing. Thanks a lot for doing this. I think it's almost ready to be merged.

My only concern (which is sort-of unrelated) is that keeping the `LC_ALL=C` prefix in front of each command could be pretty tricky to maintain down the road. I know you raised this point already and maybe we should reopen the discussion with @alex5imon on that subject.

It would be nice to build this into the `run_cmd_log` function to keep it at one place rather than having to do it manually. There are 24 places in the updater alone, and I can totally see myself forgetting to add that prefix the next time I change something.

I guess that in most cases we would want to have it there, there are going to be a few that we will want localised. Maybe we could do something like:

```python
def run_cmd_log(cmd, force_english=True):
    ...
```

Now we're at the beginning of our dev cycle so there would be plenty of time to test if we decide to change `run_cmd_*` functions.

@hzulla I'm happy to change this in your PR once we figure out whether we want to do it.

hzulla:
Adding LC_ALL is just a temporary fix.

It's easiest to fix and simplify the run_cmd issue when kano-toolset is up on github.

Then, all LC_ALL additions can be removed, again.

`force_english` isn't right, because locale C isn't the same as en_US or en_UK and may behave differently.

My suggestion is the opposite to yours: To add a default LC_ALL=C environment setting to the current run_cmd and add a new variant of run_cmd that will preserve the locale, so something like `run_cmd(cmd, locale=True)`.
Because usually, run_cmd is done to call and parse shell commands and only few of these calls are actually locale dependent (e.g. when you call Chromium for the proxy settings).

pazdera:
@hzulla: Oh, I see. Yeah, `force_english` wouldn't be quite right then.

I think we both meant the same thing with the second point though. To use `LC_ALL=C` by default in `run_cmd` and provide an optional argument to keep the output localised.

tombettany:
Merged in https://github.com/KanoComputing/kano-updater/pull/67 instead so closing.

Issue #64, opened 2015-02-19
(!) uses different colour than the text next to it


hzulla:
Hi,

[bang.png](https://github.com/KanoComputing/kano-updater/blob/master/kano_updater_gui/media/images/bang.png) is partially transparent white and thus brightens up the background colour, the text next to it is [CSS lightgray](https://github.com/KanoComputing/kano-updater/blob/766249cafb1483a3bee3e703f83a8149c92e371b/kano_updater_gui/media/css/updater.css#L23) (#D3D3D3). This results in a slight difference in colours, depending on the background.

![badbang](https://cloud.githubusercontent.com/assets/1705654/6267807/91a23808-b847-11e4-88e4-79ef2cd46d21.png)

This requires a new version of bang.png without partial transparency.

tombettany:
@TommySaehl Are you able to provide what @hzulla needs?

TommySaehl:
Yeah, sure! 
But, we're not able to change the color of the label so it matches 'bang.png' on the screens? 
Will look better than gray, but isn't vital if its complicated.

hzulla:
Well, the text isn't transparent and [this old, possibly outdated article says Gtk doesn't do transparent labels](http://stackoverflow.com/questions/2869169/setting-opacity-on-a-pygtk-label), but the article also suggests a solution to simulate the effect. The easiest solution is to use a nontransparent (!), though.

TommySaehl:
Oh, sorry but that isn't what I meant. We could just change the colour to #E8B8B8 for example, and that would make it look like this;

![image](https://cloud.githubusercontent.com/assets/1634790/6270487/abf808be-b85a-11e4-8ad1-f0b05d0a310e.png)
And do this on the four other screens as well. 

Totally your call @hzulla . If you need a new 'bang.png', I'll get one for you

hzulla:
Oh. So obvious. Will do that (won't be back with my RPi till Monday, though).

PR #66, opened 2015-02-24
fix colours for public service announcement next to (!) bang.png

hzulla:
This is a fix for #64


tombettany:
Tested this and I think that it looks right (it is hard to tell but I think that it is good). Merging. Thanks @hzulla for this fix.

hzulla:
Thanks.

PR #69, opened 2015-03-10
Add addtional python modules

carolineclark:


carolineclark:
@pazdera @tombettany here is a pip module I need for the sentry server, if you could add it to your changes, would be super useful.

PR #71, opened 2015-03-11
Update pip

pazdera:
Importing pip locally and changing the `install-pip` to `update-pip`, because it doesn't make much sense to run pip install docopt and then install-pip.

@tombettany


tombettany:
I'm okay with this. Merging.

pazdera:
@tombettany Thanks!

PR #78, opened 2015-03-12
Remove the old `kano_updater_gui` package

tombettany:
This is unnecessary in the new updater scheme as it is now a sub-package
of `kano_updater`.


pazdera:
:+1:

PR #79, opened 2015-03-12
Remove check-for-updates entry in the sudoers file

tombettany:
check-for-updates no longer exists so it is unnecessary to keep the
entry in the sudoers file.


pazdera:
:+1: Very good catch ...

PR #84, opened 2015-03-19
UI: Periodically check that install is running

tombettany:
If the install thread should die unexpectedly then the UI will just
hang. Prevent this behaviour by polling to see if the install thread is
still alive.

cc @pazdera


pazdera:
Looks good, merging ...

PR #90, opened 2015-03-25
Update i18n hooks

hzulla:
This adds some missing i18n hooks for the new updater.


hzulla:
It was fairly easy to test a local version of kano-updater-gui, but how do I testrun the new version? Thanks.

tombettany:
Hmm, that is a tricky one - short of actually running it on a system with `apt` I'm not sure. @pazdera Any ideas on this one?

tombettany:
Good work by the way @hzulla. Sorry for the trouble of having to redo the translations.

kano_updater/commands/install.py
```diff
@@ -36,7 +36,7 @@ def install(progress=None):
     if status.state == UpdaterStatus.INSTALLING_UPDATES:
         msg = 'The install is already running'
         logger.warn(msg)
-        progress.abort(_(msg))
+        progress.abort(_('The install is already running'))
```
tombettany: @hzulla We definitely need to ensure that all the logging happens in English (to give us a hope of debugging it!) but this clearly isn't the ideal situation - having to define the strings twice. Do you know of any way of telling the logger to use the raw string but then telling `progress` to use the translated version?

I'm happy with what you have done here (and is indeed what I would do with my current knowledge of the translation engine) but in an ideal world we wouldn't need repeat the strings and only present the translated version when it hits the user.

Any ideas from anyone would be greatly appreciated @Ealdwulf @alex5imon @skarbat @pazdera @carolineclark @convolu

hzulla:
Looks like this is a case for deferred translation using N_(). See https://docs.python.org/2/library/gettext.html for a brief description of the concept.

Ealdwulf:
Suppose we define N_ to return both translated and untranslated:

def N_(x):
  return (x, _(x))

Then we could modify the logger to detect a tuple and:
* store both the translated and untranslated message, so we and our use can
both read it
* return the translated message

Then we would be able to do:
          trans=logger.warn(N_('The install is already running'))

     progress.abort(trans)

hzulla:
Sounds good.

hzulla:
The more I think about it...

The logger is for internal / debugging use and it's totally OK if it logs English messages, only. By the time Kano users look under the hood and into json log files, they'll be ready and willing to look up English words to understand what's going on.

Making it accept tuples and then return one part of the tuple is a bit of an unexpected behaviour to allow this particular rare use.

So forgive me for retracting my previous comment, but maybe logger should continue to log English only and stay simple to use and without side effects.

Ealdwulf:
Hanno,
You're probably right. Also it would require everything to require a new
rev of kano-toolset.

Alex

tombettany:
I'm with you @hzulla. In https://docs.python.org/2/library/gettext.html#deferred-translations it appears that we just need to use `N_(` around the strings when we define them and then log the message and translate the output

```
msg = N_('Some text to log and display')
logger.warn(msg)
progress.abort(_(msg))
```

I'm not sure why, in the example on the webpage, it has
```
def N_(message): return message
```
but maybe that is just so the code runs without `gettext.install`.

However, we do need to account for 
```
you will need to teach your message extraction program to look for translatable strings marked with N_(). pygettext and xpot both support this through the use of command line switches.
```

hzulla:
My Pi will be back on Tuesday, then I can update this patch to use N_().

alex5imon:
sounds good!

hzulla:
Aaaaargh.

The more one looks at the code, the more one gets bitten by unicode conversion frustrations ([as described here](https://pythonhosted.org/kitchen/unicode-frustrations.html)).

Ok, taking a step back: There are various places in kano python code where it writes stuff to the console, to files or to the Gtk GUI. However, with Python 2.7, one has to decide where the conversion from/to Unicode has to take place.

For the Gtk GUI, it's easy. Its API accepts Unicode, so conversion happens when you call Gtk GUI stuff and it's easy to add translations for its message strings. That's why my earlier PRs didn't trip over this problem.

But see [here](https://github.com/hzulla/kano-updater/blob/d642d47e8524c35ff6a76484cdb0f464f1f41eaa/bin/kano-updater#L91). That line breaks breaks for the format string `"{}: {}"`, which should be `u"{}: {}"` instead, but then `enforce_root()` is the next one to break at [stderr output](https://github.com/KanoComputing/kano-toolset/blob/c5eeb566d488d47738cf4c2e3bee364f4350a8f2/kano/utils.py#L339).

So I guess, kano-toolset needs to be unicode-safe, first.

Then, one needs to decide about console output and translations. kano-wifi needs to be translated, as it's a user-facing text user interface. But what about the console output of kano-updater?

So I guess this PR here is premature and kano-toolset needs to be worked on, first.

alex5imon:
@hzulla I think you are right, we need to short out `kano-toolset`.

Regarding `kano-wifi`, don't worry about this tool because we are changing it to a GUI shape. So no more console anymore.

hzulla:
I guess this PR can be put on hold. kano-toolset needs to be unicode-safe, first (in a way that doesn't break ASCII-only use of kano-toolset).

hzulla:
@alex5imon just curious and only half kidding: Is it worth going the pain or are you planning to move to Python 3 anytime soon?

alex5imon:
I don't think so @hzulla , there are quite a few compatibility issues with Python3 and our current codebase.

PR #91, opened 2015-04-02
Don't remove the pidfile in early exit case

Ealdwulf:
I think this fixes the extra popups.


pazdera:
Thanks @Ealdwulf! I'll merge it right away and rebuild the package :-).

PR #92, opened 2015-04-07
Installing requests security module, to remove further warnings

skarbat:
* This installation should remove further complains from pip:
   "InsecurePlatformWarning: A true SSLContext object is not available"


Ealdwulf:
That looks pretty safe :-)

PR #93, opened 2015-04-08
Fixing pip requests installation

skarbat:
* Square brackets name a dependency package,
   so removing the additional line to install "requests",
   because pip complains with an error, saying it is covered
   by the previous line.

cc @Ealdwulf

### DIFF
--- a/python_modules
+++ b/python_modules
@@ -1,7 +1,6 @@
 pip
 ping
 requests[security]
-requests
 tzupdate
 pam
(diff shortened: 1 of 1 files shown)


skarbat:
Does that look right @pazdera @tombettany ?

pazdera:
Oh, I didn't know that!

Yeah, looks good to me :-). Please merge :shipit:

PR #94, opened 2015-04-09
Sanity check for status json file on the widget, and the updater

skarbat:


tombettany:
Looks good @skarbat :+1:

PR #95, opened 2015-04-09
Aborting PIP install command chain if network fails

skarbat:


kano_updater/commands/install.py
```diff
@@ -225,7 +225,10 @@ def install_pip_packages(progress):
...
         success = run_pip_command("install --upgrade {}".format(pkg))
 
-        # TODO: abort the install?
         if not success:
             msg = "Installing the '{}' pip package failed".format(pkg)
             logger.error(msg)
+            if not is_internet():
+                msg="Network is down, aborting PIP install"
+                logger.error(msg)
+                raise IOError(msg)
```
tombettany: Hmm, do we want to end everything - if they don't have internet but have already downloaded the debian packages then these can be installed. Maybe we should skip.

Also, in theory, because we cache the downloads in [commands/download.py](https://github.com/KanoComputing/kano-updater/blob/master/kano_updater/commands/download.py#L133-L157) there should be a way of just installing the cached version without needing internet; I realise that the operation that we do here does require internet but the entire install process shouldn't and if we can find a way of not requiring it here then that would be ideal.

Ealdwulf: @tombettany Maybe I'm not getting the context, but don't we basically have
to assume that the debian packages depend on the pip packages? It which
case it might not be a good idea to install them. I agree with the rest of
your comment though.

skarbat: Good point @tombettany . It is possible to tell pip to install in 2 stages `pip install --download /var/somewhere/pips` followed by `pip install --no-index --find-links=file:///var/somewhere/pips`. Skipping for now, but sharing this for later implementation.

tombettany: @Ealdwulf You make an interesting point. You are right, sometimes we do depend on the pip packages. The majority of the time these are just updates to those packages rather than adding new ones but you're right, we might be in a situation where we add a pip package and depend on it - maybe we should abort install.

PR #96, opened 2015-04-17
check-gui: Raise instead of keeping above

pazdera:
To avoid interference with minecraft.

Related to: https://github.com/KanoComputing/peldins/issues/1757

cc @Ealdwulf @alex5imon

### DIFF
--- a/kano_updater/ui/available_window.py
+++ b/kano_updater/ui/available_window.py
@@ -43,7 +43,7 @@ class NotificationWindow(Gtk.Window):
 
         self.set_icon_name('kano-updater')
 
-        # Make sure this window is always above
+        # Put the window above all the existing ones when it starts
         self.raise_()
 
         image = Gtk.Image()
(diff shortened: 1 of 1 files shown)


kano_updater/ui/available_window.py
```diff
@@ -44,7 +44,7 @@ def __init__(self):
         self.set_icon_name('kano-updater')
 
         # Make sure this window is always above
```
alex5imon: you might want to change the comment as well

Ealdwulf:
RelaunchWindow() also does keep_above, I assume that one is not relevant to check-gui? Otherwise looks good.

kano_updater/ui/available_window.py
pazdera: Good point Alej, thanks!

pazdera:
@Ealdwulf Relaunch window is a sort-of interim window that is shown when the updater relaunches itself during the update phase. There it's actually useful, because we want to make sure to cover everything else with that.

tombettany:
Looks fine to me

PR #97, opened 2015-04-20
Moving the interval guards outside of the check_for_updates function

pazdera:
So the updater doesn't show the dialog when there is an update from
a previous check, but it wasn't supposed to check in the first place
due to the interval guards being set.

cc @tombettany @alex5imon

### DIFF
--- a/bin/kano-updater
+++ b/bin/kano-updater
@@ -32,6 +32,7 @@ import os
 import sys
 import gettext
 import atexit
+import time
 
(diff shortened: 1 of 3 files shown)


alex5imon:
@pazdera looks ok to me

@skarbat ?

skarbat:
looks good to me @pazdera @alex5imon

alex5imon:
Please merge :+1:

PR #98, opened 2015-04-23
Freezing the installation of python requests library version

skarbat:
Not entirely sure but this could potentially fix the json problem encountered, Radek? @pazdera


pazdera:
Very good idea, thanks Albert! Merging right away.

PR #99, opened 2015-04-28
Create a random character on update to 2.0

convolu:
Generate an initial character for users that update

### DIFF
--- a/kano_updater/scenarios.py
+++ b/kano_updater/scenarios.py
@@ -278,6 +278,6 @@ class PostUpdate(Scenarios):
 
     def beta_133_to_beta_134(self):
         pass
-    
+
     def beta_134_to_beta_200(self):
         pass
--- a/kano_updater/scenarios.py
+++ b/kano_updater/scenarios.py
@@ -281,9 +281,9 @@ class PostUpdate(Scenarios):
         pass
 
     def beta_134_to_beta_200(self):
-        if is_installed('kano-character-cli'):
-            cmd = 'kano-character-cli -c "Judoka_Base" "Hair_Black" "Skin_Orange" -r -s'
-            run_cmd_log(cmd)
-        else:
-            logger.warn(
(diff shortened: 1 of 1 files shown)


convolu:
@pazdera @alex5imon Could you please review?

kano_updater/scenarios.py
```diff
@@ -278,6 +279,11 @@ def beta_132_to_beta_133(self):
...
         pass
-    
+
     def beta_134_to_beta_200(self):
-        pass
+        if is_installed('kano-character-cli'):
+            cmd = 'kano-character-cli -c "Judoka_Base" "Hair_Black" "Skin_Orange" -r -s'
+            run_cmd_log(cmd)
+        else:
+            logger.warn(
```
alex5imon: Should this just be a warning? Shouldn't we try to fix it by installing it (and any other dependencies)?

convolu: This should happen in the Post Update step. If `kano-character-cli` is missing, it probably means that there is something wrong with the update. Should I bump this to error?

alex5imon: Definitely to error but you might also want to fix the problem

pazdera: @alex5imon I think that the role of the check is that if the update fails really badly, this won't make it worse. If you don't have the `kano-profile` installed that usually means that your install is pretty badly damaged.

convolu: Updated in https://github.com/KanoComputing/kano-updater/commit/a8117e1da1d87407c3ef99f0d857b449d9a6ad3e

convolu:
Are these changes ok, may I merge?

pazdera:
@convolu Looks good to me!

alex5imon:
Please merge :+1:

convolu:
Thanks, merging

PR #100, opened 2015-04-28
Increasing the font size of the info text shown during installation

pazdera:
cc @alex5imon @tombettany


alex5imon:
Please merge :+1:

PR #102, opened 2015-04-29
bin: Adding the first-boot subcommand

pazdera:
This should be used from kano-uixinit at the end of the init flow to
make sure the user isn't prompted for update during the first 24 hours
from creating his account.

Needs a change in kano-uixinit which will be ready shortly.

cc @tombettany @alex5imon


pazdera:
@Ealdwulf Oh, that's a very good point. I need to change that too. We should be checking once a week everywhere now. Thanks!

PR #103, opened 2015-05-01
Increase the check interval to 7 days

pazdera:
The update check was set to an incorrect interval.

@alex5imon It's a small change but we don't have to merge it if you don't want to.


alex5imon:
@pazdera wrap it within brackets and merge :)

PR #106, opened 2015-05-13
Adding a new step in the upgrade stage to fix homedire permissions

skarbat:
* A process will go through all home directories
   and apply KanoOS 2.0 permissions on sensitive files & dirs (tracker, logs)

### DIFF
--- a/bin/repair-homedir-permissions
+++ b/bin/repair-homedir-permissions
@@ -1,0 +1,52 @@
+#!/usr/bin/env python
+#
+#  correct-homedir-permissions
+#
+# Copyright (C) 2015 Kano Computing Ltd.
+# License: http://www.gnu.org/licenses/gpl-2.0.txt GNU General Public License v2
+#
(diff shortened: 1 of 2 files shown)


pazdera:
Looks good to me, @skarbat :-).

The only thing is that the people who have already updated to 2.0.0 and have the broken permissions will not benefit from this fix (the hook 134->200 hook won't run for them). I'm not sure whether that's what you intend to do?

skarbat:
ah! Radek @pazdera, my understanding is that this issue is only appearing for users upgrading from 1.x? But can't hurt to do it for 2.0.0 -> 2.0.1 as well (just fixed). Besides, I have implemented it in a standalone tool, so in the worst case we can tell users to `sudo correct-homedir-permissions`.

skarbat:
I am merging this one now @pazdera @Ealdwulf @alex5imon

pazdera:
:+1: Thanks!

PR #107, opened 2015-05-15
install: Make sure there's enough free space available

pazdera:
This commit makes the install command check for free space before
anything happens. If there's less than 1GB left, it won't proceed
with the install and optionally tries to expand the card.

Related to: https://github.com/KanoComputing/peldins/issues/1798

cc @alex5imon  @tombettany


tombettany:
Looks fine to me

PR #109, opened 2015-05-19
update_home_folders_from_skel: Run in a try

pazdera:
This will stop the updater from freezing and log the traceback, should there be a problem in the home directory.

Related to: https://github.com/KanoComputing/peldins/issues/1820

cc @alex5imon @tombettany @skarbat @Ealdwulf


skarbat:
Good spotting @pazdera makes sense the freeze was here.

tombettany:
Looks good.

PR #110, opened 2015-05-20
Updated list of Python modules using i18n literals

skarbat:
cc @pazdera @tombettany


tombettany:
@skarbat, good idea but I don't know if this does anything but increases the number of files which are scanned when creating the `po` files for the translators. I don't think that this fixes the `_` not defined problem but I'm happy with this anyway.

skarbat:
Thanks @tombettany, merging then.

PR #111, opened 2015-05-20
Cleanup

tombettany:
- Adds missing logger import
 - Fixes raising NotImplemented which isn't an exception (use NotImplementedError)
 - Fixes `unexpectantly` misspelling
 - Fix some line length issues
 - Other assorted improvements

Note: Does not fix the '`_` referenced before assignment' problem

cc @pazdera


pazdera:
Looks good to me :-) :shipit:

PR #112, opened 2015-05-20
Using _ for an unused variable conflicts with gettext

pazdera:
This should fix it

cc @tombettany @skarbat


tombettany:
Looks fine :shipit:

PR #113, opened 2015-05-21
LXPanel Plugin: Move images into dedicated directory

tombettany:
For ease of browsing and modification, move the LXPanel plugin images
into their own directory

cc @pazdera


pazdera:
Looks good! :-) Thanks for doing that :shipit:

PR #115, opened 2015-05-21
Post update

tombettany:
* Adds screen after updating
 * Tracking

cc @pazdera


pazdera:
Can't see any obvious problems, :shipit:

PR #116, opened 2015-05-22
Adding a runtime dependency on lxpanel

pazdera:
This should fix the updater fail with lxpanel.

cc @tombettany @alex5imon


tombettany:
Great spot @Ealdwulf. This is fine. Merging now.

PR #118, opened 2015-05-22
Froze futures module version

carolineclark:
The Sentry module seems to be the only package that needs the futures module, so we're freezing the version.  Later versions of the futures modules seem to break the sentry module, which breaks the Ultimate Parental Control.


convolu:
Looks ok to me

PR #120, opened 2015-05-22
boot: Relaunch the updater when it's interrupted

pazdera:
cc @tombettany


tombettany:
Beautiful code @pazdera :stuck_out_tongue_winking_eye:. :shipit:'n it.

PR #122, opened 2015-06-24
Added dialog if no updates available

alex5imon:
https://github.com/KanoComputing/peldins/issues/1885

### DIFF
--- a/lxpanel-plugin/kano_updater.c
+++ b/lxpanel-plugin/kano_updater.c
@@ -32,13 +32,13 @@
 
 /* Icon file paths */
 #define NO_UPDATES_ICON_FILE \
-	"/usr/share/kano-updater/images/widget-no-updates.png"
+    "/usr/share/kano-updater/images/widget-no-updates.png"
 #define UPDATES_AVAILABLE_ICON_FILE \
-	"/usr/share/kano-updater/images/widget-updates-available.png"
+    "/usr/share/kano-updater/images/widget-updates-available.png"
 #define DOWNLOADING_UPDATES_ICON_FILE \
-	"/usr/share/kano-updater/images/widget-downloading-updates.png"
+    "/usr/share/kano-updater/images/widget-downloading-updates.png"
 #define UPDATES_DOWNLOADED_ICON_FILE \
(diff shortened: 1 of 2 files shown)


alex5imon:
@pazdera please review

pazdera:
@alex5imon This might be a bit tricky. I don't think that the launch_cmd function blocks until the command has finished. So there could be a race.

alex5imon:
@pazdera changed to `system` instead of `launch_cmd`, good call!

lxpanel-plugin/kano_updater.c
```diff
@@ -380,71 +382,79 @@ void install_clicked(GtkWidget *widget, gpointer data)
...
 
-void check_for_updates_clicked(GtkWidget *widget,
-			       kano_updater_plugin_t *plugin_data)
+void check_for_updates_clicked(GtkWidget *widget, kano_updater_plugin_t *plugin_data)
 {
-	launch_cmd(CHECK_FOR_UPDATES_CMD,  "kano-updater");
+    // launch_cmd can't be used because we need to wait for 
+    // the command to finish
+    kdesk_hourglass_start("kano-dialog");
+    system(CHECK_FOR_UPDATES_CMD);
```
tombettany: My only concern here is exactly the reason that `system` was used rather than `launch_cmd` - because it blocks. Would this freeze the UI? If so, could this be run in a thread instead?

alex5imon: @tombettany a bit sceptic of using threads in this case. The UI frozen is the menu, not big deal. If you could test it in today's image and let me know what you think that would help

tombettany:
This also doesn't merge directly - I suspect that it is because of all the tabs becoming spaces.

We currently have chosen a coding standard for Python but none for C/C++. I propose that we adopt one to avoid these troubles. Any ideas @alex5imon @skarbat @Ealdwulf @pazdera @convolu @carolineclark @rluz?

alex5imon:
@tombettany very similar rules should apply to C++, hence no tabs, 4 spaces for tabulation.

tombettany:
@alex5imon I don't have any problem with it now.

PR #123, opened 2015-06-25
Notifications: Suppress messages when the GUI is running

tombettany:
KanoComputing/peldins#1902
When the updater is in full-screen install mode, it will change its
status file to account for the progress made in the install. This,
however, generates notifications about the status of the process which
are visible above the GUI itself. Prevent this by providing a flag in
the status file which can disable these messages.

Integrate this flag to suppress messages when installing in full-screen
mode and reset it when a `clean` is called.

KanoComputing/peldins#1791
Also pauses all notifications until the updater has run its course so that
they are not lost but not shown during this time.

cc @pazdera

### DIFF
--- a/bin/kano-updater
+++ b/bin/kano-updater
@@ -63,9 +63,17 @@ from kano_updater.utils import make_low_prio, install_docopt, is_running, \
 def clean_up():
     remove_pid_file()
 
+    status = UpdaterStatus.get_instance()
+    status.is_gui_running = False
+    status.save()
+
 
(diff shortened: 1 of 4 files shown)


pazdera:
Looks good to me! Please merge :shipit:

Radek Pazdera
Software Engineer at Kano <http://www.kano.me>
@radekpazdera <https://twitter.com/radekpazdera>

tombettany:
Thank you Pazzy D

PR #124, opened 2015-06-26
Notifications: Rename flag for muting notifications

tombettany:
The original flag was called `is_gui_running` but this can be confusing
and suppressing the notifications may be desired in situations other
than when the GUI is running so rename it to a more sensible name
(`notifications_muted`).

cc @pazdera


pazdera:
Like the change :+1:. Feel free to merge :shipit:

tombettany:
Merging now.

PR #125, opened 2015-06-26
Flush the logs as the last step when closing

tombettany:
If the logs are flushed before the end, some of the logs won't be saved
during the update process so move them to the last of the tasks.

cc @pazdera


pazdera:
Also good :shipit:

PR #127, opened 2015-06-26
Notifications: Don't show on boot if a dialog appears

tombettany:
KanoComputing/peldins#1915
When booting, if the updater had to reboot to be able to continue, a
message is shown to prompt the user to carry on with the update. This
accompanies a status change which triggers a notification. If this is
the case, disable the notification so that two calls to action aren't
present.

cc @pazdera

### DIFF
--- a/kano_updater/commands/clean.py
+++ b/kano_updater/commands/clean.py
@@ -11,7 +11,7 @@ from kano_updater.status import UpdaterStatus
 # from kano_updater.ui.
 
 
-def clean():
+def clean(dry_run=False):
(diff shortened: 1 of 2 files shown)


pazdera:
I'm not sure where does the mute flag get cleared?

tombettany:
We `exec()` straight after so the mute flag gets cleared by the next run of the updater.

pazdera:
Right, I didn't know we relaunch from the boot screen as well. In that case, please merge :shipit:

tombettany:
Thanks Pazzy D

PR #128, opened 2015-06-26
Post-update: Add new graphic for changes window

tombettany:
KanoComputing/peldins/issues/1914
An animated version of the post-update changes window has been created
so use that instead.

cc @convolu @pazdera


pazdera:
Really like the graphic, please merge :shipit:

PR #129, opened 2015-06-26
i18n: Use the kano-i18n library for i18n init

tombettany:
Instead of each package handling i18n initialisation independently, use
the `kano-i18n` library.

cc @pazdera


pazdera:
Looks good :shipit:

PR #132, opened 2015-07-02
Notifications failsafe

pazdera:
Make sure the updater doesn't freeze without having to depend on kano-widgets

cc @tombettany @alex5imon


tombettany:
Looks good @pazdera

PR #133, opened 2015-07-02
Fix no-install

convolu:
With these changes we replace the use of the deprecated `--no-install` while downloading
the packages.

As the pip documentation now suggests we should create and maintain our own cache.
This also changes the way we install packages as we need to look in the newly created cache.

A failsafe method has been added that ensures that the packages are installed even if they are
not present in the cache


convolu:
Merging for testing

PR #136, opened 2015-09-03
Added telnet, required by use-the-force script

alex5imon:
https://github.com/KanoComputing/peldins/issues/1943


convolu:
Looks ok to me :+1: . Adding telnet in the dependencies of `kano-desktop` seems like a good idea as well, see [this comment](https://github.com/KanoComputing/kano-desktop/pull/108#issuecomment-137789918)

PR #137, opened 2015-09-15
Add tracking for successful hotfix installation

convolu:


convolu:
@tombettany @radujipa This should add tracking after a successful installation of a hotfix.
Merging right away into our feature branch.

PR #140, opened 2015-09-28
Use the format X.Y.Z-B for the package version no

convolu:
The X.Y format of the version may cause issues and goes against the assumptions about urgent updates. This rectifies this

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,4 @@
-kano-updater (2.2-1) unstable; urgency=low
+kano-updater (2.2.0-1) unstable; urgency=low
 
   * Add support for security hotfixes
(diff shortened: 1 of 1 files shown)


convolu:
@skarbat Do you think this will fix the dependency issue?

skarbat:
I think that should fix it @costas, but I am wondering if it would make more sense to fix at the other end? (kano-desktop), I think we don't use 3 digit versioning across our packages? Otherwise looks good to me.

convolu:
Indeed most of the apps don't follow this but I believe that this should be the format that we use from now on. This way it will be easier to follow urgent updaters if there are any.

The updater should set the example :-).
I am merging to restart the build

PR #144, opened 2015-10-12
Flappy judoka

radujipa:
@tombettany @pazdera  could you please have a quick look over the changes in the updater for the review?


radujipa:
Related to https://github.com/KanoComputing/engage/issues/66

kano_updater/ui/install_window.py
```diff
@@ -31,7 +32,11 @@ def __init__(self):
         apply_styling_to_screen(self.CSS_FILE)
 
         Gtk.Window.__init__(self)
-        self.fullscreen()
+        # self.fullscreen()
+        screen = Gdk.Screen.get_default()
+        width = screen.get_width()
+        height = screen.get_height()
+        self.set_size_request(width, height + 80)
```
tombettany: Is this for the Gtk fullscreen hack? If so, can we have a comment here to explain and does the height really need to be `+ 80` can't you just add `1`?

kano_updater/ui/install_window.py
```diff
@@ -45,6 +50,9 @@ def __init__(self):
         self.show_all()
         self._set_wait_cursor()
 
+        self._install_screen._pgl.hide()
```
tombettany: A bit cheeky to use private methods of objects outside the object but if there is no other way...

kano_updater/ui/install_window.py
radujipa: Yup, that was the Gtk hack to make the windows behave and listen to `set_keep_below(True)`. The `+80` value is to make sure the updater top app bar is above the screen and not showing. Will add a comment.

kano_updater/ui/install_window.py
```diff
@@ -119,9 +150,16 @@ def _no_updates(self):
...
-    def update_progress(self, percent, msg, sub_msg=''):
+    def update_progress(self, percent, msg, phase_name, sub_msg=''):
         self._install_screen.update_progress(percent, msg, sub_msg)
 
+        # enable flappy-judoka only for the RPI2
+        # enabling flappy-judoka launch only after these phases (when a reboot is iminent)
+        if is_model_2_b():
+            if phase_name in ['downloading', 'downloading-pip-pkgs', 'init', 'installing-urgent']:
+                self.get_toplevel().connect('key-release-event', self._launch_game)
+                self._install_screen._pgl.show()
```
tombettany: Private member use again.

kano_updater/ui/install_window.py
```diff
@@ -54,6 +62,29 @@ def __init__(self):
 
         self._start_install()
 
+    def _launch_game(self, window=None, event=None):
+        try:
+            if event and event.get_keycode()[1] == 44:  # [J] key
```
pazdera: You can use a Gdk constant here instead: `Gdk.KEY_j`

https://lazka.github.io/pgi-docs/index.html#Gdk-3.0/constants.html#Gdk.KEY_j

kano_updater/ui/install_window.py
```diff
@@ -54,6 +62,29 @@ def __init__(self):
...
+                    self._first_game_launch()
+                else:
+                    self._relaunch_game()
+        except:
+            pass
+
+    def _first_game_launch(self):
+        self.set_keep_below(True)
+        os.system('{} &'.format(FLAPPY_PATH))
+        os.system('kill -9 `pgrep -f kano-feedback`')      # kill kano-feedback-widget
```
pazdera: It might be slightly more efficient to use `pkill -f -9`. That wouldn't launch two processes.

pazdera:
@radujipa Why is the whole game with all the code in the `media/` directory of the updater? Wouldn't it be easier to make a separate repo and package for it?

radujipa:
@pazdera it was @alex5imon 's decision, however the game folder is in such a way that making a deb package/repo out of it in the future would be easy enough

kano_updater/ui/install_window.py
radujipa: Ah, I've looked for this and did not find it! Thank you so much, will put it in now! :)

kano_updater/ui/install_window.py
radujipa: Well spotted, thank you @tombettany! I forgot I did it the dirty way, when initially I had everything in `views/install.py` which is much better! I'll move things there which will make things cleaner in `install_window.py`. I will however need to call the `hide` and `show` methods on the label, because at the moment I do not know whether I can add more things to the `progress_grid` after creation/showing. I'll leave it as a TODO.

kano_updater/ui/install_window.py
pazdera: I think this get_keyval should return the code that will correspond with the constant instead of the HW scan code: https://lazka.github.io/pgi-docs/index.html#Gdk-3.0/unions/Event.html#Gdk.Event.get_keyval

radujipa:
@tom @pazdera implemented suggested refactors

tombettany:
@radujipa Looks reasonable to me.

radujipa:
:shipit:

PR #148, opened 2015-10-19
First boot countdown

radujipa:
Added the capability of setting a time slot after the first ever boot in which update
notifications are supressed. The time slot is currently three days.

WARNING: This is still susceptible to date/time changes after the first boot!

@alex5imon @tombettany @pazdera could you please review this?

### DIFF
--- a/kano_updater/status.py
+++ b/kano_updater/status.py
@@ -56,6 +56,7 @@ class UpdaterStatus(object):
         self._last_check = 0
         self._last_check_urgent = 0
         self._last_update = 0
+        self._first_boot_countdown = 0
         self._is_urgent = False
         self._is_scheduled = False
         self._notifications_muted = False
@@ -77,6 +78,7 @@ class UpdaterStatus(object):
                 data['last_update']
                 data['last_check']
                 data['last_check_urgent']
+                data['first_boot_countdown']
                 data['is_urgent']
                 data['is_scheduled']
                 data['is_shutdown']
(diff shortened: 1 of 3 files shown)


kano_updater/status.py
```diff
@@ -190,6 +194,19 @@ def last_check_urgent(self, value):
...
         self._last_check_urgent = value
 
+    # -- first_boot_countdown - used to stop updates for a set amount of time
+    @property
+    def first_boot_countdown(self):
+        return self._first_boot_countdown
+
+    @first_boot_countdown.setter
+    def first_boot_countdown(self, value):
+        if not isinstance(value, int):
```
pazdera: A slightly more _pythonic_ (yeah, it's a horrible word) to do this would be:

```
try:
   self._first_boot_countdown = int(value)
except ValueError:
  ...
```

pazdera:
Looks good to me! :shipit:

kano_updater/status.py
```diff
@@ -77,6 +78,7 @@ def load(self):
                 data['last_update']
                 data['last_check']
                 data['last_check_urgent']
+                data['first_boot_countdown']
```
tombettany: Given that we are adding this `first_boot_countdown` field, are we sure that this won't ever load the old version of the file and fail?

radujipa: Yup, the field was added just like when we added `last_check_urgent`, `is_shutdown` etc.
It worked fine for me. Did you test it?

tombettany:
Then I'm happy @radujipa :shipit:

radujipa:
thanks @pazdera @tombettany 
Merging and rebasing `jessie` branch

PR #150, opened 2015-10-20
Added tracking to Flappy Judoka

radujipa:
Tracking events added:
  - session length: how long the game is being run for
  - action: when the game is being launched
  - data: player score when flappy dies

@alex5imon @pazdera could you please review this?

### DIFF
--- a/kano_updater/ui/media/flappy-judoka/bin/flappy-judoka
+++ b/kano_updater/ui/media/flappy-judoka/bin/flappy-judoka
@@ -44,11 +44,19 @@ if __name__ == '__main__' and __package__ is None:
 from src.core.gameloop import Gameloop
 from src.utils import debugger, is_running
 
+from kano_profile.tracker import track_action, session_start, session_end
+
+
+# how long the game is being played for
+tracking_session = None
+
 
 # UNCOMMENT THE 2 LINES BELOW TO GET A PROFILE FOR THE GAME
 # from profilehooks import profile
 # @profile
 def main():
+    global tracking_session
+
     # make sure the game launches only once
     if is_running():
         debugger('flappy-judoka: main: Game is already running! Exiting..')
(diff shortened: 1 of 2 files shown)


kano_updater/ui/media/flappy-judoka/src/core/gamestate.py
```diff
@@ -383,6 +385,14 @@ def __init__(self, pipes, score):
 
         self.pipes.reset_pipes()
 
+        try:
+            # TODO: track best score as well
+            track_data('updater-flappy-judoka', {
```
pazdera: I'd call the event 'updater-flappy-judoka-score'. Because there can be more data events per app.

radujipa:
Screenshot of the events after a few games played

screen shot 2015-10-20 at 17 29 08

kano_updater/ui/media/flappy-judoka/src/core/gamestate.py
radujipa: Ah, that was my bad (we talked about this), thank you good catch!

kano_updater/ui/media/flappy-judoka/bin/flappy-judoka
```diff
@@ -76,6 +91,12 @@ def main():
     debugger('flappy-judoka: main: Exiting')
     pygame.quit()
 
+    # letting the tracking session know the game is closing
+    try:
+        session_end(tracking_session)
+    except:
+        debugger('ERROR: flappy-judoka: main: Ending tracking session failed!')
```
pazdera: Does the debugger log it somewhere? It'd be good to get logs if this will be failing.

I got burned a few times with ignoring all exceptions -- it makes it really difficult to find the error if there's one.

radujipa: You are right, this is a more particular case so I'll use the `logger` we have.

PR #151, opened 2015-10-23
Added postupdate scenario for awarding the Computer Commander badge

carolineclark:


alex5imon:
please merge :+1:

kano_updater/scenarios.py
```diff
@@ -332,3 +332,12 @@ def beta_201_to_beta_210(self):
...
     def beta_210_to_beta_220(self):
         install('telnet python-serial')
+        try:
+            from kano_profile.apps import save_app_state_variable
+
+            save_app_state_variable("init-flow-completed", "level", 1)
+
+        except Exception as e:
+            logger.error(
+                "kano-profile could not award character creator, exception e".format(e)
```
convolu: You missed the `{}` in the error string to replace the exception message

carolineclark: Yes I did. Thank you.

kano_updater/scenarios.py
```diff
@@ -332,3 +332,12 @@ def beta_201_to_beta_210(self):
 
     def beta_210_to_beta_220(self):
         install('telnet python-serial')
+        try:
+            from kano_profile.apps import save_app_state_variable
+
+            save_app_state_variable("init-flow-completed", "level", 1)
```
convolu: This will only happen for the user who is running the updater, not for all of the installed users in the system

tombettany:
Looks fine to me.

convolu:
:shipit:

@tombettany Are the notifications disabled during the postupdate hooks?

tombettany:
@convolu [save_app_state_variable_all_users()](https://github.com/KanoComputing/kano-profile/blob/9bea5b3bc6d3f35137ab55bd10d25dc9dca3446b/kano_profile/apps.py#L159-L178) never calls [save_app_state_with_dialog()](https://github.com/KanoComputing/kano-profile/blob/a3a4ae378d088d842d2dae27d894e73d2fc27f5d/kano_profile/badges.py#L225-L274) so no GUI should appear anyway. Otherwise, I think that the notifications are just paused so if there were to be one appear then it would happen after reboot.

PR #152, opened 2015-10-27
Upload tracking data on update

pazdera:
Adding a new aux task.

cc @radujipa @tombettany @alex5imon

### DIFF
--- a/kano_updater/auxiliary_tasks.py
+++ b/kano_updater/auxiliary_tasks.py
@@ -55,4 +55,4 @@ def _expand_rootfs():
     run_cmd_log('/usr/bin/expand-rootfs')
 
 def _sync():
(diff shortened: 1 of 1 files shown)


alex5imon:
Is it worth doing here another `kano-sync --skip-kdesk --sync --backup --upload-tracking-data -s` ?

pazdera:
Yeah, we could do the whole sync there. I'll add it.

radujipa:
I tried testing this, however the Updater fails during the downloading phase due to broken packages..

To me, the PR looks good, since it's using `run_cmd_log` so we'll get logs and it's run in a subprocess so probably try/except is not really needed here, correct?

pazdera:
@radujipa That should be correct.

radujipa:
:shipit:

PR #153, opened 2015-10-27
Relaunch: Fix passing splash PID between instances

tombettany:
When relaunching the updater, a relaunch splash screen is created to
mask the underlying updater restart. This relies on passing the splash
screen PID between updater instances so that the latter version can kill
the splash screen.

The handling of both passing the PID of the splash screen to the
`exec()` call for relaunch and killing the splash screen after relaunch
were within the same function so some variables were used twice (to mean
both the PID of the launched splash and the currently running splash
PID) which lead to undetermined behaviour. Fix this by separating the
two uses.

@pazdera @radujipa could you review?


pazdera:
Looks good to me :shipit:

PR #156, opened 2015-12-16
Added a post update scenario for 220 to 230

radujipa:


radujipa:
https://github.com/KanoComputing/peldins/issues/2146

convolu:
Is this the best place for this or can it be added to the debian package postinst script?

radujipa:
@Ealdwulf Good point! I've added the line break.

@convolu That is a good question! @pazdera @alex5imon ?

alex5imon:
@convolu @radujipa if we are doing `os.system('sudo adduser {} i2c'.format(get_user_unsudoed()))
 ` in Peldins (so the new version 2.3.0 and onwards already has it) then it can be done on the Updater.

skarbat:
@convolu @radujipa I think it also needs the user to belong to the `i2c` group?

```
albert@albert ~ $ ls -l /dev/i2*
crw-rw---- 1 root i2c 89, 1 Jan  5 21:17 /dev/i2c-1
albert@albert ~ $ id
uid=1001(albert) gid=1001(albert) groups=1001(albert),4(adm),5(tty),20(dialout),24(cdrom),27(sudo),29(audio),44(video),46(plugdev),60(games),100(users),101(input),1000(kanousers)
```

kano_updater/scenarios.py
```diff
@@ -341,3 +344,24 @@ def beta_210_to_beta_220(self):
             logger.error(
                 "Could not award Computer Commander badge, import error"
             )
+
+    def beta_220_to_beta_230(self):
+        os.system('sudo adduser {} i2c'.format(get_user_unsudoed()))
```
convolu: We probably need all of the linux users on the kit to be added to this group and change it so new users are added as well.

pazdera:
@radujipa If you think there might be two or more independent packages that might need this in the future, doing it system-wide like this is easier. If not and it's related solely to that package, it makes more sense to do it in the package.

skarbat:
good point @pazdera , perhaps `kano-init adduser` would be interested to know about it,

 * https://github.com/KanoComputing/kano-init/blob/jessie/kano_init/user.py#L24-L25

And kano-greeter too...

 * https://github.com/KanoComputing/kano-greeter/blob/jessie/bin/kano-greeter-account#L72-L73

cc @radujipa @convolu

radujipa:
@pazdera @skarbat Very good spot! Thank you! I've added the `i2c` group [to the greeter](https://github.com/KanoComputing/kano-greeter/pull/26) and [to kano-init](https://github.com/KanoComputing/kano-init/pull/16).

kano_updater/scenarios.py
```diff
@@ -359,4 +359,22 @@ def beta_210_to_beta_220(self):
             )
 
     def beta_220_to_beta_230(self):
-        pass
+        os.system('sudo adduser {} i2c'.format(get_user_unsudoed()))
```
convolu: Would it be possible to apply this over all normal users currently on the system? We want all of our previously created users to be able to use the speaker led deamon without sudo.

Ealdwulf: @radujipa @convolu Shoudln't this be os.system('sudo usermod -a -G i2c {}'.format(get_user_unsudoed()))? Adduser adds an actual user.

convolu: You are right @Ealdwulf !

radujipa: It looks like `kano-greeter` and `kano-init` use something similar [here](https://github.com/KanoComputing/kano-init/blob/master/kano_init/user.py#L141) and [here](https://github.com/KanoComputing/kano-greeter/blob/master/bin/kano-greeter-account#L96). Good catch, will have it changed.

tombettany:
Rebase it

radujipa:
@tombettany done

tombettany:
:+1:

PR #157, opened 2016-01-07
Remove modemmanager package on 2.3.0

convolu:
This PR removes the `modemmanager` package during the preupdate scenario to OS version 2.3.0.

As there is a minor chance that a user manually installed the package, we first check that is was installed automatically (which btw would also be the case if the user needs it from a different package that he has installed)

### DIFF
--- a/kano_updater/scenarios.py
+++ b/kano_updater/scenarios.py
@@ -132,6 +132,9 @@ class PreUpdate(Scenarios):
         self.add_scenario("Kanux-Beta-2.1.0", "Kanux-Beta-2.2.0",
                           self.beta_210_to_beta_220)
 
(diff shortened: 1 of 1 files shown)


convolu:
@skarbat @pazdera @Ealdwulf Could you please review?

skarbat:
looks great to me @convolu !

pazdera:
Looks good, but it might conflict with Radu's scenario PR.

convolu:
@pazdera good catch.

@radujipa This may conflict with https://github.com/KanoComputing/kano-updater/pull/156 . That one doesn't seem complete so I'll merge this and help with the merging :-D

PR #158, opened 2016-01-13
220->230: Install rsync during update

pazdera:
Related to: https://github.com/KanoComputing/peldins/issues/2216

cc @skarbat @alex5imon

### DIFF
--- a/kano_updater/scenarios.py
+++ b/kano_updater/scenarios.py
@@ -359,4 +359,4 @@ class PostUpdate(Scenarios):
             )
 
     def beta_220_to_beta_230(self):
-        pass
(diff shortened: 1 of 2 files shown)


skarbat:
Looks good @pazdera , makes me think... should we say `no-install-recommends` as well here?

 * https://github.com/KanoComputing/kano-updater/blob/jessie/kano_updater/utils.py#L210-L211

pazdera:
That's a good point, @skarbat.

pazdera:
@skarbat Added it.

skarbat:
great Radek! @pazdera

PR #159, opened 2016-01-20
Install Make Light on 2.3.0

convolu:
towards https://github.com/KanoComputing/peldins/issues/2162

### DIFF
--- a/kano_updater/scenarios.py
+++ b/kano_updater/scenarios.py
@@ -360,37 +360,64 @@ class PostUpdate(Scenarios):
             )
 
     def beta_220_to_beta_230(self):
+        # A few helper fns to keep the scenario tidy
+        def add_users_to_i2c_group():
+            try:
+                linux_users = get_users()
+                if not linux_users:
+                    logger.error('beta_220_to_beta_230: linux_users is empty!')
+
+                for user in linux_users:
+                    os.system('sudo usermod -a -G i2c {}'.format(user))
+
+            except Exception as e:
+                logger.error(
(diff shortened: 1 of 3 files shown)


convolu:
@tombettany @pazdera @radujipa Could you please review?

radujipa:
:shipit:

kano_updater/scenarios.py
```diff
@@ -361,6 +361,7 @@ def beta_210_to_beta_220(self):
 
     def beta_220_to_beta_230(self):
         install('rsync')
+        run_cmd_log('kano-apps install --no-gui powerup')
```
pazdera: I think you need sudo here since the install is non-interactive. It won't be able to prompt you for a password.

convolu: We've done in this way before https://github.com/KanoComputing/kano-updater/blob/jessie/kano_updater/scenarios.py#L287-L288 and https://github.com/KanoComputing/kano-updater/blob/jessie/kano_updater/scenarios.py#L327

I could change it if you want, just to be sure

pazdera: @convolu No, it's correct. I started typing before thinking :-P

convolu:
@tombettany @radujipa @pazdera I have reorganised the post update scenario. What do you guys think?

radujipa:
The sign that a big one is coming :shipit:

tombettany:
Looks good @convolu

convolu:
I've updated as per your comments, @tombettany. I'll merge this to get time to test it.

PR #162, opened 2016-01-22
Adding hover to the lxpanel widget

pazdera:
Related to: https://github.com/KanoComputing/peldins/issues/2183

~~The icons need transparency change before merging:~~

~~all `widget-*` in https://github.com/KanoComputing/kano-updater/tree/master/lxpanel-plugin/images~~

cc @alex5imon @skarbat @neue


skarbat:
looks ok to me @pazdera

neue:
I don't have write access to this repo, so here they are:

widget-downloading-updates
widget-no-updates
widget-updates-available
widget-updates-downloaded

radujipa:
cc @alex5imon

alex5imon:
@neue @radujipa not sure we can add icons for the hover effect. It just uses the default effect

pazdera:
Brilliant, thanks @neue!

@alex5imon You're right. The problem was that the icon's we have there now have a white border around them which becomes visible with the new hover effect.

pazdera:
Committed the icons, this should be now ready to merge.

PR #163, opened 2016-02-02
Removing the brain reference from the updater

pazdera:
We removed the brain reference from the kano-apps dialog. Removing it from here as well.

What do you guys think? @neue @radujipa @alex5imon


radujipa:
Much better. However, with `expand-rootfs` now running at shutdown, they will only see this in `2.3.0` if they update on the first boot, right?

pazdera:
@radujipa Yep, very very edge-case scenario. But since it's already there ...

radujipa:
:shipit:

PR #164, opened 2016-02-02
Set a different state on reboot

pazdera:
Related to: https://github.com/KanoComputing/peldins/issues/2257

cc @radujipa @convolu @tombettany


pazdera:
Just finished testing this change and it does fix the problem.

radujipa:
Not an expert here, but it looks ok to me :+1:

PR #166, opened 2016-02-03
check-for-app-updates: New aux task

pazdera:
This will check whether we updated any of the app icons and allow users to redownload them.

cc @tombettany @alex5imon @convolu


tombettany:
:+1:

PR #167, opened 2016-02-04
Relaunching Lxpanel in the cleanup if FlappyJudoka was launched

radujipa:
This is supposed to act as a safe guard when the Updater fails for whatever reason and FlappyJudoka was launched. Previously, several processes were killed in order to keep the game on top of the window. This is no longer needed and if the updater crashes, the OS will not be in an 'unstable' state anymore.

@tombettany @pazdera @convolu

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,7 +1,12 @@
+kano-updater (2.4.0-1) unstable; urgency=low
+
+  * Improved Flappy Judoka game launching
+
+ -- Team Kano <dev@kano.me>  Wed, 20 Jan 2016 14:35:00 +0000
+
 kano-updater (2.3.0-1) unstable; urgency=low
 
   * Install Make Light by default
-  * Improved Flappy Judoka game launching
 
  -- Team Kano <dev@kano.me>  Wed, 20 Jan 2016 14:35:00 +0000
 
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,6 +1,7 @@
 kano-updater (2.3.0-1) unstable; urgency=low
 
   * Install Make Light by default
+  * Improved Flappy Judoka game launching
 
  -- Team Kano <dev@kano.me>  Wed, 20 Jan 2016 14:35:00 +0000
 
--- a/kano_updater/ui/install_window.py
+++ b/kano_updater/ui/install_window.py
@@ -1,11 +1,10 @@
-
 # install_window.py
 #
-# Copyright (C) 2015 Kano Computing Ltd.
+# Copyright (C) 2015-2016 Kano Computing Ltd.
 # License: http://www.gnu.org/licenses/gpl-2.0.txt GNU GPL v2
 #
 # Installer main window
-#
+
 
 import os
 from gi.repository import Gtk, Gdk, GLib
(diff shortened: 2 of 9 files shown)


kano_updater/utils.py
```diff
@@ -124,6 +124,11 @@ def kill_apps():
         variables))
 
 
+def relaunch_lxpanel():
+    cmd = '/usr/bin/lxpanel --profile LXDE &'
+    run_bg(cmd)
```
tombettany: When the updater exits, will it take `lxpanel` with it?

kano_updater/utils.py
```diff
@@ -124,6 +124,11 @@ def kill_apps():
         variables))
 
 
+def relaunch_lxpanel():
+    cmd = '/usr/bin/lxpanel --profile LXDE &'
```
alex5imon: Should we check if lxpanel is already running first?

radujipa: Fixed this.

kano_updater/utils.py
radujipa: Quick answer is: no. :)

radujipa:
I didn't manage to get this quite working yet. For some reason `kano-world-widget` and `kano-feedback-widget` would not launch. 

`OSError: [Errno 13] Permission denied: '/root/.kanoprofile'` in `".../kano/utils/file_operations.py", line 150`

radujipa:
Related to https://github.com/KanoComputing/peldins/issues/2279

radujipa:
I've refactored with a new solution based on `wmctl` thanks to @Ealdwulf which does not need to kill any widgets. The only drawback now is that any dialog launched will put the game behind the updater while it is still capturing mouse and keyboard input (hence the `kill_flappy_judoka` calls).

@alex5imon @tombettany @convolu @pazdera

radujipa:
Finally got this working properly. 

@Ealdwulf @tombettany could you please review?

debian/changelog
```diff
@@ -1,6 +1,7 @@
 kano-updater (2.3.0-1) unstable; urgency=low
 
   * Install Make Light by default
+  * Improved Flappy Judoka game launching
```
tombettany: As this is being released for `v2.4.0`, you're going to need to bump the version here.

kano_updater/ui/install_window.py
```diff
@@ -31,13 +30,7 @@ def __init__(self):
...
 
         Gtk.Window.__init__(self)
-        # self.fullscreen()
-        # Gtk hack: set the width, height of the window to larger than screen
-        # resolution in order fix set_keep_below(True) which doesn't work with fullscreen
-        screen = Gdk.Screen.get_default()
-        width = screen.get_width()
-        height = screen.get_height()
-        self.set_size_request(width, height + 80)
+        self.fullscreen()
```
tombettany: This works now?

debian/changelog
radujipa: Ah right! Good catch! I started this back in `2.3.0` :)

kano_updater/ui/install_window.py
radujipa: Yup, reverted back to what it was before I added Flappy.

kano_updater/ui/progress.py
```diff
@@ -24,10 +24,12 @@ def _change(self, phase, msg):
                       phase.get_main_phase().label, phase.get_main_phase().name, msg)
 
     def _error(self, phase, msg):
+        kill_flappy_judoka()
```
tombettany: Don't you already do this in the `self._window.error()` function. Regardless, this progress object shouldn't contain any information about the details of the GUI - it should only be used to connect the backend with the GUI.

radujipa: I added `kill_flappy_judoka` in quite a few other places. Is this the only redundant one?

tombettany: Probably not but this was the most obvious one.

radujipa:
If there are no other suggestions/worries on this, I will merge EOD.

PR #168, opened 2016-02-05
Deps: Bump kano-toolset dependency

tombettany:
The PostUpdate scenarios require the use of `open_locked`'s `timeout`
parameter but if `kano-toolset` hasn't been updated to include this
feature before the Updater runs the cached version of the `kano.utils`
module is used and hence fails. Resolve this by bumping the dependency
requirement.

@Ealdwulf @pazdera @convolu do you think that this will fix the caching issue?

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -8,7 +8,7 @@ Build-Depends: debhelper (>= 9), build-essential, pkg-config, libgtk2.0-dev,
 
 Package: kano-updater
 Architecture: any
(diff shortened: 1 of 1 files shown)


convolu:
it will definitely help, let's try it.

tombettany:
Okay, merging now so that we can get started on the lengthy testing.

pazdera:
:+1:

PR #169, opened 2016-02-05
Relaunch: Ensure that update proceeds after relaunch

tombettany:
During the relaunch of the updater, somehow the state gets reset to
`NO_UPDATES` which causes the relaunch to exit prematurely. Prevent this
from happening by forcing the state to be the value that it should be
after a relaunch.

@pazdera @convolu @radujipa to review


pazdera:
Looks good, let's test

radujipa:
Nice one!

tombettany:
Updater should work after this fix so merging now.

pazdera:
:shipit: Woot!

PR #171, opened 2016-02-19
Removed dialog asking to continue update after reboot

radujipa:
Related to https://github.com/KanoComputing/peldins/issues/2067, but this one is optional.


tombettany:
So we want to continue regardless? If this is the desired behaviour then I would agree with the implementation but why are we doing it automatically?

radujipa:
The main reason behind it is because since the user already agreed to update, we should finish the task rather than asking them again.

cc @alex5imon

alex5imon:
I'm ok with it

tombettany:
Fair enough. I'm okay with this PR then :+1:

PR #175, opened 2016-03-09
Separate update

Ealdwulf:
NB this isn't ready to merge, but I'd like people's opinion on whether this is the right approach to installing the package @tombettany @alex5imon @skarbat @radujipa 

This PR:
* Adds a list of packages in 'special_packages.py' which don't trigger an OS update.
*  Introduces a new updater command 'kano-updater update_ind_pkg <pkg>'. This enters a new state  'INSTALLING_INDEPENDENT'  and installs the package, then transitions to UPDATES_INSTALLED. The updater is prevented from going into install mode when an independent package is being installed and vice versa.
* Adds a new command 'avail_ind_pkgs' which lists the spcial packages which are available to be updated. This may be removed if we decide on a different way of communicating with kano-dashboard.

Questions:
* is UPDATES_INSTALLED the correct state after 'INSTALLING_INDEPENDENT'? I'm not quite sure what effect it has.
* Is the progress object relevent to update_ind_pkg?

### DIFF
--- a/bin/kano-updater
+++ b/bin/kano-updater
@@ -17,6 +17,8 @@ Usage:
   kano-updater set-scheduled (1|0)
   kano-updater first-boot
   kano-updater clean
+  kano-updater avail-ind-pkgs
+  kano-updater update-ind-pkg <package>
   kano-updater ui (relaunch-splash <parent-pid> | boot-window | shutdown-window)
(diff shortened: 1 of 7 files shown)


tombettany:
:+1:

Ealdwulf:
Thanks @tombettany , merging...

PR #176, opened 2016-03-30
Update status.updatable_independent_packages after installing an update

Ealdwulf:
This PR fixes a problem noticed by @tombettany in that the status file wasn't being updated after installing an independent package, thus causing the gui to think it could still be installed.

While looking at this I noticed that because the list of updatable independent packages is made using whatever priority level the check command is called  with, that field in the status file coudl flip flop between having and not having the package.  Therefore I have looked through the use of priority and made a couple of changes:

Generally, only deal with independent packages *independently* at ordinary priority . At URGENT priority,  treat them the same as other packages (IE< if an urgent update is pushed to an 'idependent' package, this is installed using the urgent update facility).
Detail:
* Always make the list of updatable independent packages using Priority.NONE
* When checking for urgent updates, don't exclude independent packages.
@radujipa @tombettany

### DIFF
--- a/kano_updater/apt_wrapper.py
+++ b/kano_updater/apt_wrapper.py
@@ -208,7 +208,8 @@ class AptWrapper(object):
 
     def is_update_available(self, priority=Priority.STANDARD):
         for pkg in self._cache:
-            if pkg.name not in independent_install_list:
+            # exclude independent packages, UNLESS this is an urgent update
+            if priority == Priority.URGENT or pkg.name not in independent_install_list:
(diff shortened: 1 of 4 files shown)


tombettany:
@Ealdwulf Should it also reset the `state` field to `no-updates` so that it is clean for the next run of the updater?

Ealdwulf:
@tombettany  - I think this is ready to merge...

tombettany:
:+1:

Ealdwulf:
Thanks @tombettany , merging...

PR #177, opened 2016-04-11
Fixed Jessie urgent updates into APT during an update

skarbat:
* On Jessie, the APT url was not being updated, so no urgent updates were possible.
 * Wheezy users should also get the urgent updates (not strictly necessary though, I think).

@Ealdwulf


alex5imon:
Nice one @skarbat !!

skarbat:
Thanks @alex5imon ! merging for testing tomorrow.

PR #178, opened 2016-04-11
Speed up kano-updater ui boot-window

Ealdwulf:
This PR adds fast script for querying updater status file and running kano-updater if necessary.
A dependency on jq (json query) is introduced.

@tombettany @skarbat @alex5imon  
Note that there is some maintenance cost to this as extra triggers must be added in both places.
For https://github.com/KanoComputing/peldins/issues/2327


alex5imon:
👍

skarbat:
Nice one @Ealdwulf

Ealdwulf:
Thanks, merging...

PR #179, opened 2016-04-13
Add fix to set `dtparam=audio=on` in /boot/config.txt

Ealdwulf:
This should add the setting from https://github.com/KanoComputing/kano-settings/pull/369 for users who update.
Function copied from spi modification in 2.3.0 scenario.
@alex5imon @skarbat


alex5imon:
Please merge 👍

PR #180, opened 2016-04-18
Bump version number for 3.0.1

Ealdwulf:
@skarbat @alex5imon @tombettany

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,9 @@
+kano-updater (3.0.1-1) unstable; urgency=low
+
+  * Bump version
+
+ -- Team Kano <dev@kano.me>  Mon, 18 Apr 2016 11:23:00 +0100
+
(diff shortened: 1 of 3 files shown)


radujipa:
Don't forget the changes in peldins as well to go with this 👍

tombettany:
👍

Ealdwulf:
Thanks guys, merging...

skarbat:
:+1:

convolu:
A minor point here regarding the urgent updates. I'm not sure whether the logic has evolved since, but in the initial implementation we kept the last part of the version to 0, only to be bumped for packages in need of an urgent update. So the assumption was that the only thing that was bumped was the X and the Y in the X.Y.Z scheme.

Ealdwulf:
Thanks Costas. I'll change it to 3.1.0

alex5imon:
@Ealdwulf @convolu thanks guys!

PR #183, opened 2016-04-28
Notifications: Disable available and downloaded notifications

tombettany:
KanoComputing/os-dashboard#141
The dashboard handles downloading the updates in the background
ungracefully so to combat this don't show the notifications which permit
this action.

cc @Ealdwulf @alex5imon


Ealdwulf:
Looks good to me :+1:

PR #185, opened 2016-06-01
Add uart config for make-light

Ealdwulf:
This PR adds post-update scenario to add 'enable_uart=1' to config.txt. Some complication is required to be compatible with older versions of kano_settings. I have tested both (by deliberatily breaking kano-settings so the exception case happens).

For https://github.com/KanoComputing/peldins/issues/2369
@tombettany @alex5imon


alex5imon:
This is great @Ealdwulf . Please make sure it goes into the changelog

Ealdwulf:
Thanks @alex5imon, merging.

PR #187, opened 2016-06-07
Skipping Overworld onboarding and creating flag for Dashboard

radujipa:
Related to https://github.com/KanoComputing/kano-overworld/issues/504

cc @Ealdwulf @tombettany

### DIFF
--- a/kano_updater/scenarios.py
+++ b/kano_updater/scenarios.py
@@ -11,7 +11,7 @@ from kano.logging import logger
 
 from kano_updater.os_version import OSVersion, TARGET_VERSION
 from kano_updater.utils import install, remove_user_files, update_failed, \
-    purge, rclocal_executable, migrate_repository, get_users
+    purge, rclocal_executable, migrate_repository, get_users, run_for_every_user
 from kano.utils import run_cmd_log, get_user_unsudoed, write_file_contents, \
     is_installed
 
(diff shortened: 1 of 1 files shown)


kano_updater/scenarios.py
```diff
@@ -494,6 +494,12 @@ def enable_audio_device():
                 logger.error("end_config_transaction not present - update to kano-settings failed?")
         enable_audio_device()
 
+        # tell kano-overworld to skip onboarding stage
+        run_cmd_log('/usr/bin/luajit /usr/share/kano-overworld/bin/skip-onboarding.lua', unsudo=True)
+
+        # tell dashboard to skip Overworld and kit setup onboarding phase
+        run_cmd_log('touch ~/.dashboard-onboarding-vid-only', unsudo=True)
```
tombettany: This will need to use [`run_for_every_user()`](https://github.com/KanoComputing/kano-updater/blob/17f595026e5dfd6e35bf86bcf92d2714a6a3bc11/kano_updater/utils.py#L84-L86) instead as this will only touch the file for the user performing the update, not everyone else.

Also, can we change the name of the file to something along the lines of `~/.dashboard-click-onboarding-done` so that it makes sense for a new user.

radujipa: Good spot @tombettany! All current users on the kit at the time of the update should skip the onboarding directly to the video. Sure, I'll change the filename as well for you.

skarbat: There's also this helper:

 * https://github.com/KanoComputing/kano-updater/blob/jessie/kano_updater/utils.py#L404

tombettany:
The dashboard stuff looks fine to me. @Ealdwulf to check the overworld bits.

Ealdwulf:
Makes sense to me 👍

PR #189, opened 2016-06-08
Independent Pkgs: Update the sources before installing

tombettany:
KanoComputing/peldins#2375
When installing the independent updates, it is possible for the source
repositories to have changed in the time between the package being
identified as having updates available and the user actually choosing to
update it. Resolve this problem by updating the APT sources immediately
before attempting to update the package.

cc @Ealdwulf @pazdera


pazdera:
Looks good to me :+1:

Ealdwulf:
👍

PR #190, opened 2016-06-09
Pip Download: Switch to modern "download" command

tombettany:
KanoComputing/peldins#2380
The pip download was using the `pip install --download` command during
the download phase. This has been deprecated and will be removed in pip
10 so switch to use its replacement: `pip download`.

cc @Ealdwulf @pazdera


Ealdwulf:
👍

PR #191, opened 2016-06-09
Push audio dither setting to updating users.

Ealdwulf:
For https://github.com/KanoComputing/peldins/issues/2384,
if we decide we like the audio dither, we can push it to updating users too.
@tombettany @alex5imon


tombettany:
Implementation looks good.

tombettany:
Merging so that we can test - we can always revert.

PR #195, opened 2016-07-20
[DO NOT MERGE] Additional option to not send control notifications

skarbat:
* This option is needed when downloading updates in the background
   while the screen saver is running.
 * In this case the desktop icon hooks are in charge for
   disabling notifications

### DIFF
--- a/bin/kano-updater
+++ b/bin/kano-updater
@@ -11,7 +11,7 @@ kano-updater will help you keep your Kano OS up-to-date.
 
 Usage:
   kano-updater check [--gui] [--interval <time>] [--urgent]
-  kano-updater download [--low-prio]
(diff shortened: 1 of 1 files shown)


Ealdwulf:
@skarbat this hasbeen marked 'do not merge' for a while. Do we need it? I'm going to close it now because it's pointing at jessie, but not delete the branch, so feel free to reopen pointing at master.

skarbat:
I think it was a nice to have at the time which we decided to hold for the release.
It should be already covered here:

 * https://github.com/KanoComputing/kano-desktop/blob/master/kdesk/icon-hooks.sh#L88-L91

So, good to discard this PR. @Ealdwulf @radujipa

PR #196, opened 2016-07-21
Force unattended installation of bluez package

skarbat:
* During upgrade from 2.3.0 to 2.4.0, it would potentially stall asking for confirmation

@radujipa @tombettany


tombettany:
Merged with https://github.com/KanoComputing/kano-updater/pull/197

PR #197, opened 2016-07-21
Bluez: Move bluez preinstall to happen with every update

tombettany:
https://trello.com/c/1QrdXt2l/92-updater-pi2-update-from-2-4-0-to-3-4-0-failed
The update of `bluez` will often fail due to its postinstall
configuration script exiting with error status when the user does not
have a bluetooth capable device present. A work-around to this was
utilised for the update from v2.3.0 - v2.4.0 via a preupdate scenario
which installed the package manually. Unfortunately, this same problem
affects users updating from v2.4.0 to later versions so make the
solution more broad to do the install independently for every update.

This has a side-effect of adding a function where pre- and post-update
scripts which should run on every update can be located.

cc @Ealdwulf @radujipa @skarbat

### DIFF
--- a/kano_updater/scenarios.py
+++ b/kano_updater/scenarios.py
@@ -250,8 +250,7 @@ class PreUpdate(Scenarios):
     def _finalise(self):
         # When bluez is installed through a dependency it fails to configure
         # Get around this by installing it first
-        run_cmd_log('apt-get install bluez')
-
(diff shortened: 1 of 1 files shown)


tombettany:
@Ealdwulf Looks good to me

skarbat:
looks good to me too

PR #199, opened 2016-09-01
I18n prep

keberlin:


kano_updater/apt_progress_wrapper.py
```diff
@@ -62,18 +62,16 @@ def __init__(self, updater_progress, ops=[]):
         self._updater_progress = updater_progress
         self._phase_name = updater_progress.get_current_phase().name
 
-        ops = [self._get_op_key(op) for op in ops]
+        ops = [self._get_op_key(op[0]) for op in ops]
 
-        phases = [Phase(op, op) for op in ops]
+        phases = [Phase(op[0], op[1]) for op in ops]
```
tombettany: I think that this line isn't right because `ops` has now been overwritten so it is no longer a list of tuples. Either change the above to

```python
 ops = [(self._get_op_key(op[0]), op[1]) for op in ops]
```

or rename the variable and use some sort of `zip()` operation.

kano_updater/commands/install.py
```diff
@@ -313,34 +313,34 @@ def install_standard(progress, status):
...
     progress.start('preupdate')
     try:
         preup.run()
     except Exception as err:
-        logger.error('The pre-update scenarios failed.')
-        logger.error(str(err))
-        progress.abort(_('The pre-update tasks failed.'))
+        logger.error("The pre-update scenarios failed.")
+        logger.error(err.encode('utf-8'))
+        progress.abort(_("The pre-update tasks failed."))
```
radujipa: It looks like the `abort` methods here are going to do a call to the logger with translated strings.

kano_updater/ui/media/flappy-judoka/src/graphics/display.py
```diff
@@ -38,23 +38,23 @@ class Display(object):
     WIDTH = 0.60
     HEIGHT = 0.72
 
-    GAME_TITLE = 'Flappy Judoka'
+    GAME_TITLE = _("Flappy Judoka")
```
radujipa: Hmm, not sure we want to translate the name of the apps / games.

kano_updater/ui/media/flappy-judoka/src/graphics/hud.py
```diff
@@ -52,11 +52,11 @@ def __init__(self, display):
         self.font_path = os.path.join(fonts_path, 'zx_spectrum-7_bold.ttf')
         self.setup_fonts()
 
-        self.controls_hint_text = self.hint_renderer.render('[SPACE] or [UP] to Jump', 1, self.WHITE)
+        self.controls_hint_text = self.hint_renderer.render(_("[SPACE] or [UP] to Jump"), 1, self.WHITE)
         self.controls_hint_text_x = self.display_width * self.CONTROLS_HINT_X
         self.controls_hint_text_y = self.display_height * self.CONTROLS_HINT_Y
 
-        self.quit_hint_text = self.hint_renderer.render('[ESC] or [Q] to Quit', 1, self.WHITE)
+        self.quit_hint_text = self.hint_renderer.render(_("[ESC] or [Q] to Quit"), 1, self.WHITE)
```
radujipa: Not sure if I've spotted where Flappy is "installing" `kano-i18n` and the translation functions. Do you see the strings with the wide font under `en_QQ`?

keberlin:
Hi Tom,

Can you let me know where the Flappy app is started from? I'll get the i18n
init stuff added..

Keith

tombettany:
@keberlin It should be [here](https://github.com/KanoComputing/kano-updater/blob/jessie/kano_updater/ui/views/install.py#L141)

skarbat:
I have tested this PR on Spanish and English images, and seems to be working fine now.
Merging so we can test further @keberlin @tombettany @radujipa

PR #210, opened 2016-12-02
Post Update: Install numptyphysics and adventure on update

tombettany:
Add numptyphysics and adventure to the list of default apps.


tombettany:
Merging for testing.

PR #211, opened 2016-12-14
Live expand

Ealdwulf:
Not for merging yet, just putting this here before I lose it.

### DIFF
--- a/bin/expand-rootfs
+++ b/bin/expand-rootfs
@@ -63,41 +63,9 @@ p
 w
 EOF
 
-    # now set up an init.d script
-    # which will resize the filesystem on boot
-cat <<\EOF > /etc/init.d/resize2fs_once &&
-#!/bin/sh
-### BEGIN INIT INFO
-# Provides:          resize2fs_once
-# Required-Start:    $local_fs
-# Required-Stop:
-# Default-Start: 2 3 4 5 S
-# Default-Stop:
-# Short-Description: Resize the root filesystem to fill partition
-# Description:
-### END INIT INFO
-
-. /lib/lsb/init-functions
-
-case "$1" in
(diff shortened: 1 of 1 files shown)


Ealdwulf:
This is now tested and seems to work @tombettany @skarbat

bin/expand-rootfs
```diff
@@ -63,41 +63,9 @@ p
...
-    echo "Usage: $0 start" >&2
-    exit 3
-    ;;
-esac
-EOF
-
-  logger -i "registering a file system resize for next boot sequence"
-  chmod +x /etc/init.d/resize2fs_once &&
-  update-rc.d resize2fs_once defaults
+  partprobe /dev/mmcblk0 &&
```
skarbat: @Ealdwulf  Will this work if kano boots from an external disk? think automated upgrade tests framework.

skarbat:
other than comment above, looks like a wonderful improvement. Nice one, Alex!

bin/expand-rootfs
Ealdwulf: Hmm. No it won't, but I don't think it will anyway, no? It's refering to `/dev/mmcblk0` everywhere. If we want to fix that, I think it needs  a new PR.

Ealdwulf: In fact above it says 'logger -i "/dev/disk/by-label/rootfs is not an SD card. Don't know how to expand"'

skarbat:
mmm interesting. I am pretty sure I did the expansion from a usb drive, but can't remember how.
I'll keep in mind to address on a separate changeset, as you say @Ealdwulf - thanks!

Ealdwulf:
Oops forgot to merge this.

PR #212, opened 2016-12-14
Increase disk space required and fix logging errors

tombettany:
Space: Increase disk space requirement to 1.5GB

With the update to Kano OS 3.8.0, many additional apps were added which
requires more disk space to be installed. Modify the check for free disk
space to require 1.5GB, rather than the 1GB required before.

Debian: Require latest i18n versions of kano-toolset

Logging of errors can go awry if the version of kano-toolset which the
updater is started with (after updating itself) doesn't have the latest
i18n tools. This can cause logging of i18n related errors rather than
the original cause, making it very difficult to debug. Resolve this by
depending on the latest versions.

cc @radujipa @skarbat


tombettany:
Tested and looks to work okay. Merging immediately for Megan to fully test.

PR #215, opened 2016-12-14
Fixes for lost automatic login after upgrade

skarbat:
This change should fix the automatic tty login being lost, after upgrading from several older OS versions.

@tombettany @Ealdwulf


skarbat:
Thanks @Ealdwulf ! A second look helped me spot a typo, function wasn't being called.
Merging for testing.

PR #220, opened 2017-02-21
i18n: Localise the LXPanel plugin

tombettany:
Apply localisation to the Updater LXPanel plugin.

cc @skarbat @radujipa


tombettany:
[Add translation to lxpanel widgets](https://trello.com/c/bQNtp6HJ/86-add-translation-to-lxpanel-widgets)

skarbat:
👍

lxpanel-plugin/kano_updater.c
```diff
@@ -6,6 +6,8 @@
  *
  */
 
+#define GETTEXT_PACKAGE "kano-updater"
```
radujipa: I can see this was defined on line 14, duplicated?

po/Makefile
```diff
@@ -27,6 +27,8 @@ xgettext -f PYPOTFILES -L Python --force-po --from-code=utf-8 -o messages.pot \
     --package-name=$(APPNAME) --copyright-holder=$(ORG)
 xgettext -f CPOTFILES -L C -k_ -kN_ -j -o messages.pot \
     --package-name=$(APPNAME) --copyright-holder=$(ORG)
+xgettext -o messages.pot -j -L C --keyword=Q_ --keyword=N_ ../lxpanel-plugin/*.c  \
```
radujipa: So are we using `Q_` here or am I missing something?

lxpanel-plugin/kano_updater.c
tombettany: Yup, I'm an idiot

po/Makefile
tombettany: Yeah, the Gtk Internationalisation page [recommends it over `_()`](https://developer.gnome.org/glib/stable/glib-I18N.html#Q-:CAPS):

> This has the advantage that the string can be adorned with a prefix to guarantee uniqueness and provide context to the translator.

Don't really know what it means but if I did then it wouldn't be Gtk!

PR #221, opened 2017-02-22
Extent OS version with Name and Variant

radujipa:
Related to https://trello.com/c/8HLyQFpt/87-add-tracking-event-that-lets-us-know-a-kit-is-in-argentina

This one needs careful reviewing @tombettany @Ealdwulf @skarbat


radujipa:
[Add tracking event that lets us know a kit is in argentina](https://trello.com/c/8HLyQFpt/87-add-tracking-event-that-lets-us-know-a-kit-is-in-argentina)

kano_updater/os_version.py
```diff
@@ -13,55 +15,83 @@
...
+        self._name = name
+        self._variant = variant
 
         try:
             self._major_version = '.'.join(version.split('.')[0:2])
         except:
             self._major_version = version
 
     def to_issue(self):
-        return "{} {} {} \\l".format(self._os, self._codename, self._number)
```
Ealdwulf: Can this be called on old version files? Since it's in the updater.
Maybe it would be more robust to keep this additional information in a separate file, and not change the format of the existing file.

kano_updater/os_version.py
```diff
@@ -13,55 +15,83 @@
...
+        vstr = "{} {} {}".format(self._os, self._devstage, self._number)
+        # Add the name and variant to the string only if they exist.
+        vstr += ' ' + self._name if self._name else ''
```
radujipa: @Ealdwulf To your point from above, yes the function can be called on the old file, hence why here I append and empty string when the `name` cannot be found.

I was careful to preserve backwards compatibility here.

Ealdwulf: AH, you're doing it there. Okay :-)

PR #222, opened 2017-03-02
Updated scenario coverage of versions with 'Lovelace' OS name

radujipa:
Related to https://trello.com/c/qGwjC9zc/161-rc-updater-gets-stuck-updating-from-v3-8-0-version-is-too-old-you-need-to-reflash

@tombettany @skarbat


skarbat:
I agree with Tom's comment - it's also likely `Lovelace` will become something else pretty soon?

radujipa:
I would do it this way given that stripping it away would involve some more invasive changes to [os_version.py](https://github.com/KanoComputing/kano-updater/blob/master/kano_updater/os_version.py)...

tombettany:
Hmm, yeah. Okay. Looking at the file it looks like this PR's approach might well be the safest way.

skarbat:
I see your point @radujipa ... I can't find any other mention of the version name on the kit.
Perhaps we can go forward and we can add a QA test scenario for that. 👍

radujipa:
I've just finished running through and update from 3.8.0 to latest rc with this fix and everything seems in order afterwards.

radujipa:
[RC - Updater: Gets stuck updating from v3.8.0 - "[...] version is too old, you need to reflash [...]"](https://trello.com/c/qGwjC9zc/161-rc-updater-gets-stuck-updating-from-v3-8-0-version-is-too-old-you-need-to-reflash)

PR #223, opened 2017-03-02
Hidding Home button during the updating process

radujipa:
Related to https://trello.com/c/oled38vu/761-safely-hide-to-dashboard-home-button-when-the-updater-is-running

@skarbat @tombettany


radujipa:
[Safely hide 'To Dashboard' home button when the Updater is running](https://trello.com/c/oled38vu/761-safely-hide-to-dashboard-home-button-when-the-updater-is-running)

skarbat:
Looks fine @radujipa , although... are these cases ok for relaunching the updater?

* https://github.com/KanoComputing/kano-updater/blob/4d96c3594d2e5425e7f8dc2cddf27855074f9429/bin/kano-updater#L139-L149
* https://github.com/KanoComputing/kano-updater/blob/4d96c3594d2e5425e7f8dc2cddf27855074f9429/bin/kano-updater#L139-L149

radujipa:
@skarbat I would say that it covers that scenario as well, since when it relaunches it does so with `--gui` and it calls `run_install` again if it sees the `--splash-pid` arg.

But just to be sure, I will double check that tomorrow. 👍

skarbat:
👍

radujipa:
Ok, I have now tested this in a number of scenarios and it looks good :)

PR #225, opened 2017-04-19
CI: Utilise package autobuild function

tombettany:
Eliminate the knowledge from the Jenkinsfile of how to do the package
building to allow this to be centralised.

cc @skarbat


skarbat:
👍

PR #226, opened 2017-04-24
Removed ~/.asoundrc file from all user home dirs

radujipa:
cc @tombettany @skarbat

PR #227, opened 2017-04-24
Correctly enable Overture during an upgrade

skarbat:
This change applies the new systemd style in kano-init during an upgrade.
The side effect was that Overture was being launched on top of Dashboard.

Sorry, it will probably clash with: https://github.com/KanoComputing/kano-updater/pull/226

@radujipa @tombettany


skarbat:
[(3) As a user, I want to see the matrix animation during onboarding [5]](https://trello.com/c/uz6Ihbdo/888-3-as-a-user-i-want-to-see-the-matrix-animation-during-onboarding-5)

PR #229, opened 2017-05-02
Disabling any hat power buttons during Updater

radujipa:
@tombettany @skarbat


radujipa:
[(3) There is no restriction on when the shutdown menu pops up by pressing the power button on the PiHat](https://trello.com/c/nAQEjY0F/871-3-there-is-no-restriction-on-when-the-shutdown-menu-pops-up-by-pressing-the-power-button-on-the-pihat)

PR #230, opened 2017-05-03
Installed kano2-app after the update

radujipa:
@tombettany


radujipa:
[(1) As a user, I want the electron app installed on the dashboard](https://trello.com/c/OOQQiuv0/770-1-as-a-user-i-want-the-electron-app-installed-on-the-dashboard)

PR #231, opened 2017-05-03
CK2 Pro: Disable power menu while the updater is running

tombettany:
Disable the shutdown menu for the CK2 Pro power button using the same
mechanism as is done for the CK2 Lite's power button. Also wrap all the
imports in `try`s to safeguard against a version of Kano Peripherals
that doesn't support the hats.

Complements https://github.com/KanoComputing/kano-peripherals/pull/27

Based on https://github.com/KanoComputing/kano-updater/pull/229.

cc @radujipa


tombettany:
[(2) There is no restriction on when the shutdown menu pops up by pressing the power button on the PiHat](https://trello.com/c/nAQEjY0F/871-2-there-is-no-restriction-on-when-the-shutdown-menu-pops-up-by-pressing-the-power-button-on-the-pihat)

kano_updater/utils.py
```diff
@@ -642,13 +651,29 @@ def enable_power_button():
...
-    if pihat_iface:
-        pihat_iface.set_power_button_enabled(True)
+    try:
+        pihat_iface = get_pihat_interface()
+        if pihat_iface:
+            pihat_iface.set_power_button_enabled(True)
+
+        pro_hat_iface = get_ck2_pro_interface()
+        if pro_hat_iface:
+            pro_hat_iface.set_power_button_enabled(False)
```
radujipa: Boolean here needs to be `True`

tombettany: I'm an idiot

tombettany:
Thanks @radujipa, made the requested changes.

PR #232, opened 2017-05-05
Verification that the kit is plugged in before updating

radujipa:
Depends on https://github.com/KanoComputing/kano-peripherals/pull/29


radujipa:
[(3) As a CK2 Pro user, I want to be informed that updating while using the battery is DANGEROUS](https://trello.com/c/kqgJj4CH/931-3-as-a-ck2-pro-user-i-want-to-be-informed-that-updating-while-using-the-battery-is-dangerous)

debian/control
```diff
@@ -3,14 +3,33 @@ Maintainer: Team Kano <dev@kano.me>
...
+    gir1.2-gtk-3.0 (>= 3.10.2),
+    kano-settings (>= 1.3-1),
+    python-apt,
+    schedtool,
+    python-pip,
+    kano-i18n (>= 3.5-1),
+    jq,
+    kano-init,
+    kano-profile (>= 3.9.0),
+    kano-peripherals
```
tombettany: Should you not list a version of `kano-peripherals` which has the required functions? Peripherals might need a version bump.

kano_updater/utils.py
```diff
@@ -677,3 +679,58 @@ def disable_power_button():
...
+    try:
+        from kano_peripherals.wrappers.detection import is_ck2_pro
+        ck2_pro = is_ck2_pro()
+
+        from kano_peripherals.ck2_pro_hat.driver.high_level import get_ck2_pro_hat_interface
+        ck2pro_iface = get_ck2_pro_hat_interface()
+        is_battery_low = (ck2pro_iface and ck2pro_iface.is_battery_low())
+    except:
+        # Kano Peripherals doesn't support this function
+        pass
```
tombettany: Should we also set something here - the hat may exist but if we can't access the interface then we are in trouble. I'm not sure whether we should assume that the battery is fine in this situation or say that it is too risky. Presumably we should warn anyway. What do you thing @radujipa

kano_updater/utils.py
```diff
@@ -677,3 +679,58 @@ def disable_power_button():
...
+            description_text=_("Is your computer plugged in?"),
+            button_dict={
+                _("Yes"): {'color': 'green', 'return_value': True},
+                _("No"): {'color': 'red', 'return_value': False}
+            }
+        )
+        is_plugged = dialog.run()
+
+        header = ""
+        if is_battery_low:
```
tombettany: After the previous message, it is likely that the user will go and find their cable and plug it in, in which case `is_battery_low` will have changed since then so we might actually want to do the check here.

kano_updater/utils.py
```diff
@@ -677,3 +679,58 @@ def disable_power_button():
...
+
+        from kano_peripherals.ck2_pro_hat.driver.high_level import get_ck2_pro_hat_interface
+        ck2pro_iface = get_ck2_pro_hat_interface()
+        is_battery_low = (ck2pro_iface and ck2pro_iface.is_battery_low())
+    except:
+        # Kano Peripherals doesn't support this function
+        pass
+
+    if ck2_pro:
+        # Run the first dialog asking the user a question.
```
tombettany: Maybe add a comment about the fact that we can't tell the difference between people who have plugged in and those who have a high battery - I originally thought that we shouldn't bother showing this message if the battery isn't low before I realised this fact.

kano_updater/utils.py
```diff
@@ -677,3 +679,58 @@ def disable_power_button():
...
+        is_plugged = dialog.run()
+
+        header = ""
+        if is_battery_low:
+            header = _("Low Battery")
+        else:
+            header = _("Power Required")
+
+        # If the answer was negative, show another message.
+        if not is_plugged:
```
tombettany: Also, we can bail out if at this point if `ck2pro_iface.is_battery_low()` is still true (we might catch some liars here).

debian/control
radujipa: Good point, bumped

kano_updater/utils.py
radujipa: I had a little note about that in the docstring of the method :)

kano_updater/utils.py
radujipa: Good thinking, I've moved the query after the first dialog.

kano_updater/utils.py
radujipa: Yup, also incorporated that in the return value 👍

kano_updater/utils.py
radujipa: To be honest, I see this try except as a very paranoid check, Updater worthy. The only time I see this might happen is the when updating from 3.9.0, otherwise it should work fine. The functions we're calling here have logging so we'll be able to debug this if we need to.

skarbat:
Looks good @radujipa

PR #233, opened 2017-05-23
Fix overture kicking in after an upgrade

skarbat:
After an upgrade from 2.4.0, kano-init drags the overture app package,
which starts in multi-user mode. This mistakenly goes through the onboarding on reboot.
We did not care too much about this systemd mode previously.

Fixed by forcing kano-init to prepare the system for a regular graphics mode Dashboard bootup.

@tombettany @radujipa

### DIFF
--- a/kano_updater/scenarios.py
+++ b/kano_updater/scenarios.py
@@ -593,6 +593,9 @@ class PostUpdate(Scenarios):
         # tell dashboard to skip Overworld and kit setup onboarding phase
         run_for_every_user('touch ~/.dashboard-click-onboarding-done')
 
+        # tell kano-init to setup to kit to go Dashboard the systemd way
+        run_cmd_log('kano-init finalise --force')
+
     def beta_300_to_beta_310(self):
         pass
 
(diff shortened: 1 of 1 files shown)


kano_updater/scenarios.py
```diff
@@ -593,6 +593,9 @@ def enable_audio_device():
         # tell dashboard to skip Overworld and kit setup onboarding phase
         run_for_every_user('touch ~/.dashboard-click-onboarding-done')
 
+        # tell kano-init to setup to kit to go Dashboard the systemd way
+        run_cmd_log('kano-init finalise --force')
```
tombettany: Shouldn't this be in the `beta_390_to_beta_3100()` function?

skarbat: Oh you mean postupdate, very true!

kano_updater/scenarios.py
```diff
@@ -593,6 +592,9 @@ def enable_audio_device():
...
```
tombettany: Huh, how come this guy is still here? Surely we need just one of these in the the postupdate `beta_390_to_beta_3100()` function.

skarbat:
Thank you @tombettany

skarbat:
[(3) As a user I want to be able to update from 2.4.0 to latest without issues](https://trello.com/c/Sy1g16Bj/984-as-a-user-i-want-to-be-able-to-update-from-2-4-0-to-latest-without-issues)

PR #234, opened 2017-05-23
Install topmost package which contains top level packages

radujipa:
The need for a topmost package comes from the inevitable dependency
loop that emerges from the requirement that every package needs a
parent. With kano-os package acting as the topmost, the next
hierarchy level is occupied by applications. These will install libraries, 
services, 3rd party tools, etc. required by them to function.

New package is here: https://github.com/KanoComputing/kano-os

Together with: https://github.com/KanoComputing/peldins/pull/2827


radujipa:
[As a user I want to be able to update from 2.4.0 to latest without issues](https://trello.com/c/Sy1g16Bj/984-as-a-user-i-want-to-be-able-to-update-from-2-4-0-to-latest-without-issues)

PR #235, opened 2017-05-24
Fix version comparison in Pre/Post Scenarios

radujipa:
Version comparison was improved in version 3.9.0 to accommodate for
the OS name. However, when running the pre/post update scenarios
versions were being compared with their string representation rather
than using the __cmp__ in OSVersion object. This removes the str()
wrappers on versions.


radujipa:
[(P1) Updater does not run all pre/post update scenarios from v2.4.0, stops at v3.2.0](https://trello.com/c/t5X4wKyP/1002-p1-updater-does-not-run-all-pre-post-update-scenarios-from-v2-4-0-stops-at-v3-2-0)

PR #236, opened 2017-05-26
Provide a --dry-run option to query udpates availability

skarbat:
Complements this PR: https://github.com/KanoComputing/os-dashboard/pull/329

### DIFF
--- a/bin/kano-updater-quickcheck
+++ b/bin/kano-updater-quickcheck
@@ -1,3 +1,6 @@
+#!/bin/bash
+#
+# kano-updater-quickcheck
 #
 # Copyright (C) 2015 Kano Computing Ltd.
 # License: http://www.gnu.org/licenses/gpl-2.0.txt GNU GPL v2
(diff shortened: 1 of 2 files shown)


bin/kano-updater-quickcheck
```diff
@@ -5,12 +5,27 @@
...
+dry_run=$1
 
-# check if we need to run 
-DO_RUN=$(jq '.is_scheduled!=0 or .state!="no-updates"' $STATUS_FILE)
+function are_updates_available()
+{
+    # check if we need to run
+    DO_RUN=$(jq '.is_scheduled!=0 or .state!="no-updates"' $STATUS_FILE)
+    if [ "$DO_RUN" == "true" ]; then
+        return 1
```
tombettany: Hmm, I didn't think that you could return from a bash function. I thought that the best that you could do was just set a variable and check it elsewhere.

skarbat: `Functions return a value, called an exit status. This is analogous to the exit status returned by a command. The exit status may be explicitly specified by a return statement, otherwise it is the exit status of the last command in the function`

 * http://tldp.org/LDP/abs/html/complexfunct.html

You can also use a variable to return more complex things, as you say.

PR #237, opened 2017-06-12
Debug statement would prevent updates to be started

skarbat:
Code self explanatory.. sorry about that @radujipa !


skarbat:
[(P2) - Updater: When update is available when powering up, 'Loading dashboard' is displayed on top of updater window.](https://trello.com/c/dUCswERG/1005-p2-updater-when-update-is-available-when-powering-up-loading-dashboard-is-displayed-on-top-of-updater-window)

radujipa:
Tested this and looked through the Updater to see what it's used for. We should think to remove `kano-update-quickcheck` altogether.

PR #239, opened 2017-06-13
Refrain from disrupting the power button

skarbat:
Disable and re-enable the power button, only when working in GUI operation mode.


skarbat:
[(P1) shutdown-menu - Come up before Dashboard does (race condition) and takes away mouse input](https://trello.com/c/rTj4gLWJ/1070-p1-shutdown-menu-come-up-before-dashboard-does-race-condition-and-takes-away-mouse-input)

bin/kano-updater
```diff
@@ -162,19 +169,25 @@ def run_install_ind_pkg(package):
...
     # docopt wasn't installed by default prior Kano OS 1.3.3
     # It needs to be installed to make sure the updater runs
     install_docopt()
     import docopt
 
     args = docopt.docopt(__doc__, version=str(TARGET_VERSION))
 
+    gui_mode = args['--gui']
+
+    if gui_mode:
```
radujipa: We should move the hiding of the home button in there as well given the above.

skarbat: Very true! fixed

PR #243, opened 2017-06-22
Ensure kano-init dependency version and finalised after update

radujipa:
Together with the kano-init package version bump, this would
ensure any call made to kano-init will be on the latest version
of the package. We call finalise once more to be certain users
are not thrown into Overture after the update.


radujipa:
[(P1) [CARE] Onboarding runs again after 3.10 to 3.10.1 upgrade.](https://trello.com/c/ZiOh2R3Q/1113-p1-care-onboarding-runs-again-after-310-to-3101-upgrade)

tombettany:
@radujipa And a version of this one with no conflicts.

To be merged after https://github.com/KanoComputing/kano-init/pull/59

PR #246, opened 2017-07-19
Version: Bump to 3.9.1-Lovelace

tombettany:
Bump version for next release.

Needed in `master` regardless of the image generation approach.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -35,6 +35,12 @@ kano-updater (3.10.0-1) unstable; urgency=low
 
  -- Team Kano <dev@kano.me>  Tue, 21 Mar 2017 15:52:00 +0000
 
+kano-updater (3.9.1-1) unstable; urgency=low
+
+  * Bump version
+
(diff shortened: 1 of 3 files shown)


debian/changelog
```diff
@@ -35,6 +35,12 @@ kano-updater (3.10.0-1) unstable; urgency=low
 
  -- Team Kano <dev@kano.me>  Tue, 21 Mar 2017 15:52:00 +0000
 
+kano-updater (3.9.1-1) unstable; urgency=low
+
+  * Bump version
+
+ -- Team Kano <dev@kano.me>  Wed, 19 Jul 2017 16:50:00 +0000
```
radujipa: Could you edit the date of this to something in between 3.9.0 and 3.10.0 so we don't risk dpkg getting upset?

tombettany: Yeah, we had a long conversation about this bit but you are probably right, we should play it safe.

PR #248, opened 2017-08-07
Terminate gracefully when finished

skarbat:
This addition will allow to automate the updater in a sysroot.


kano_updater/ui/views/finish.py
```diff
@@ -62,3 +63,6 @@ def _finish(self, window=None, event=None):
             os.system('systemctl poweroff')
         else:
             os.system('systemctl reboot')
+
+        time.sleep(10)
+        os._exit(0)
```
tombettany: `_exit()`? Is this a thing? Or should it by `sys.exit()`?

skarbat: Oh yes. I was concerned that Gtk would not like a normal `sys.exit()`, I'll push a change and try another round. It's the only way to test it because the updater program updates its own package and relaunches itself in the process.

[os._exit()](https://docs.python.org/2/library/os.html#os._exit)

skarbat: I knew I did this for a reason... `sys.exit()` does not work, it's blindly refusing to terminate.
I could do something like `if in a sysroot then os._exit()` ...

skarbat:
Tested another round, and automated upgrades work ok now.

PR #252, opened 2017-08-25
Disable temperature and low voltage overlay warnings only

radujipa:
This change comes amid manufacturing IC specifics for the CKC
battery. As such, we disable the Raspbian overlay warning icons
only and keep the safety mechanism still active (avoid_warnings=2
disables safety as well). This complements the same work in
kano-settings and adds the change to existing users.

Complements https://github.com/KanoComputing/kano-settings/pull/453


radujipa:
[(3) As a CK2 Pro user, I don't want to be warned about low voltage levels all of the time [5]](https://trello.com/c/0SUXjR9C/1214-3-as-a-ck2-pro-user-i-dont-want-to-be-warned-about-low-voltage-levels-all-of-the-time-5)

Ealdwulf:
👍

kano_updater/scenarios.py
```diff
@@ -791,6 +791,11 @@ def beta_3_10_5_to_beta_3_11_0(self):
             # Give more current to the USB ports for the Pixel Kit.
             max_usb_current=1
             [all]
+
+            [pi3]
+            # Disable temperature and low voltage overlay warnings only, safely is still ON.
```
tombettany: Probably should be `safety is still ON.`

PR #253, opened 2017-09-04
Version: Bump to 3.12.0

tombettany:


skarbat:
Looks good to me. At some point it would be great to have a `kano-updater --dry-run` that prints out all the pre/post stages to go through, with a test to validate it.

PR #254, opened 2017-09-07
Removed local ALSA config files from all users

radujipa:
This change in a post-update from 3.11.0 to 3.12.0 is to ensure
that the local ALSA config files are removed such that only the
system ones are kept (see kano-settings, kano-desktop).


radujipa:
[[P1] As a CKC user, I do not want audio distortions when using the CKC speaker at full volume](https://trello.com/c/AnP58vC7/1331-p1-as-a-ckc-user-i-do-not-want-audio-distortions-when-using-the-ckc-speaker-at-full-volume)

PR #256, opened 2017-09-19
Replaced /boot/config.txt file with newer version

radujipa:
Numerous changes were made to the config.txt file in kano-settings
which we need align. Replacing the file rather than migrating each
change is more convenient here.


radujipa:
[(8) [CARE] Screen Resolution issue when moving from TV to monitor. [8]](https://trello.com/c/vsFlFYaY/1172-8-care-screen-resolution-issue-when-moving-from-tv-to-monitor-8)

Ealdwulf:
This means people will have to restore their audio/monitor settings. Might be a good idea to let Customer care know to expect that.

kano_updater/scenarios.py
```diff
@@ -849,3 +849,13 @@ def _add_boot_config_options(self, extra_config):
     def beta_3_11_0_to_beta_3_12_0(self):
         # Remove .asoundrc files from all users (see kano-desktop & kano-settings).
         remove_user_files(['.asoundrc'])
+
+        # Replace the config.txt after numerous changes there (see kano-settings).
+        try:
+            import shutil
+            shutil.copyfile(
```
tombettany: Can we at least backup their old file?

skarbat: I encourage some sort of backup as well

radujipa:
I've rebased this branch on the 3.13.0 image.

@tombettany @skarbat I've created a backup of the config file, good idea!
@Ealdwulf I've preserved the audio output channel after the config.txt replacing, thanks for spotting that!

skarbat:
Seen it 👍

PR #257, opened 2017-09-25
Bump OS version to 3.12.1

radujipa:


radujipa:
[Bump Kano OS version to v3.12.1](https://trello.com/c/y69bNUDq/1389-bump-kano-os-version-to-v3121)

PR #259, opened 2017-09-25
Scenarios: Set flag for CKC v1.0 users

tombettany:
Create a flag for CKC v1.0 users upgrading to 3.12.1 to allow a message
to be displayed advising on how to receive a free speaker upgrade.


kano_updater/scenarios.py
```diff
@@ -860,4 +860,18 @@ def beta_3_11_0_to_beta_3_12_0(self):
...
+            Any CKC user passing through this function must have updated and
+            hence must be a v1.0 user (CKC v1.1 ships with a higher version
+            number and has logic to determine the version by the hardware).
+
+            Add flag for these users to show a free speaker upgrade message.
+            '''
+
+            from kano_peripherals.wrappers.detection import is_ck2_pro
+            if is_ck2_pro():
+                speaker_warning_file = '~/show_speaker_warning'
```
skarbat: Mmmmm... I think this will point to `/root` as the updater is sudoed?
You have `SUDO_USER` but I'm sure we address this case on other places.

skarbat: I'm so silly... the function does it for us, sorry.

tombettany: @skarbat We run this through a `run_for_every_user()` so we should be okay.

tombettany:
[As a CKC v1.0 user, I want to see a message when I update telling me how to claim my new speaker](https://trello.com/c/WOHStdQH/1391-as-a-ckc-v10-user-i-want-to-see-a-message-when-i-update-telling-me-how-to-claim-my-new-speaker)

PR #263, opened 2017-10-16
Bumped changelog with applied changes from v3.9.2 es_AR backports

radujipa:
This is to specify that the changes made in the es_AR image need to be rebuilt, tested, and released from master as well.


radujipa:
[(5) As a user, I want the websites I visit to be simpler and faster [3]](https://trello.com/c/7kK6vXLD/1558-5-as-a-user-i-want-the-websites-i-visit-to-be-simpler-and-faster-3)

PR #264, opened 2017-10-24
Moved 'first-boot' tracking event to kano-tracker-ctl generate

radujipa:
This removes the tracking event 'first-boot' from the routine
which only sets the Updater timestamps. The responsibility of
generating this event is delegated back to `kano-ui-autostart`.

Related to: https://github.com/KanoComputing/kano-desktop/commit/babf920b7e366717bcaf89f5b7d097421b5cda42


radujipa:
[(8) As a user, I would like my kit to boot up very quickly](https://trello.com/c/kgsbu9gZ/1301-8-as-a-user-i-would-like-my-kit-to-boot-up-very-quickly)

PR #266, opened 2017-11-22
Add scenario to install scratch2.

Ealdwulf:
This PR adds a scenario which tries to insatll scratch2, making sure that the right version of rpi-chromium-mods-kano is already installed.
For https://trello.com/c/ivizQWPd


Ealdwulf:
NB, I don't attempt to remove scratch1. It seems to be harmless to leave it.

kano_updater/scenarios.py
```diff
@@ -932,4 +932,15 @@ def beta_3_12_1_to_beta_3_13_0(self):
...
+            import apt.cache
+            c = apt.cache.Cache()
+            if c['rpi-chromium-mods-kano'].installed >= '20170809':
+                install('scratch2')
+            else:
+                logger.error("beta_3_13_0_to_beta_3_14_0: wrong version of rpi-chomium-mods-kano installed: {}".format(
+                             c['rpi-chromium-mods-kano'].installed.version)
+                )
+        except:
+            logger.error("beta_3_13_0_to_beta_3_14_0: Failed to install scratch2")
```
skarbat: Should we log the exception provided there's quite a few steps in it?
As in here, for example: https://github.com/KanoComputing/kano-updater/blob/5db639c2f36a1b26f6bf76dd63f529377288de31/kano_updater/scenarios.py#L639

Ealdwulf: That's a good idea. Although, the logger has build in support for exceptions (can pass `exception=e` which I think means it extracts the traceback as well, so I'll do it that way.

PR #268, opened 2017-12-19
Improve app launch time by eliminating D-Bus iface retries

radujipa:
This sets the retry_count=0 when checking at app startup if it is
running on a CKC. It had the effect of slowing down the app launch
by ~7sec.


radujipa:
[As an engineer, I know the Updater can be launched a little quicker](https://trello.com/c/cY7Vr6W7/1701-as-an-engineer-i-know-the-updater-can-be-launched-a-little-quicker)

PR #270, opened 2017-12-20
Download & Install commands now re-checks and re-downloads over an old cache

radujipa:
### Download command now re-checks and re-downloads over an old cache

Previously, the Updater would not download newer packages in the
background once it had downloaded something once. Now, it will
check for updates again and when newer packages are found they
will be downloaded. This avoids installing an old set of packages
which could create problems over long periods of time.

### Install command now re-checks and re-downloads over an old cache

Previously, the install command simply installed what had been
already downloaded. This might create some issues when the cache
is out of date, and so before an install check for updates first,
download any new packages, and then continue with the installation.

---

**NOTE:** We should think through this carefully and spot any risky edge-cases!


radujipa:
[As a user, I do not want to risk failing an Update because of not installing the latest packages](https://trello.com/c/2EgfXT5N/1705-as-a-user-i-do-not-want-to-risk-failing-an-update-because-of-not-installing-the-latest-packages)

kano_updater/commands/download.py
```diff
@@ -55,10 +55,29 @@ def download(progress=None, gui=True):
...
+        )
+        progress.start('checking')
+        check_for_updates(progress=progress)
+        if status.state == UpdaterStatus.NO_UPDATES:
+            err_msg = N_("Latest updates have been downloaded already")
+            logger.info(err_msg)
+            progress.abort(_(err_msg))
+            return True
+
+        progress.start('downloading')
```
tombettany: It looks like this section is just the same as the `if status.state == UpdaterStatus.NO_UPDATES` case above with the exception of what it does if after checking there are no updates, would it be clearer to merge these two cases to

```python
if status.state in [UpdaterStatus.NO_UPDATES, UpdaterStatus.UPDATES_DOWNLOADED]:

    ...

    pre_check_state = status.state
    check_for_updates(progress=progress)

    if status.state == UpdaterStatus.NO_UPDATES:
        if pre_check_state == UpdaterStatus.NO_UPDATES:
            # No updates case
            return False
        if pre_check_state == UpdaterStatus.UPDATES_DOWNLOADED:
            # Updates downloaded case
            return True

    ...

elif status.state == UpdaterStatus.DOWNLOADING_UPDATES:
    ...
```

Maybe it wouldn't, but just a thought.

kano_updater/commands/install.py
```diff
@@ -74,29 +74,29 @@ def install(progress=None, gui=True):
...
+        Phase(
+            'download',
+            _("Downloading updates"),
+            40,
+            is_main=True
+        ),
+        Phase(
+            'install',
+            _("Installing updates"),
+            60,
```
tombettany: You might want this 60/40 split based on whether some updates were already downloaded because if they were then it is likely that the download isn't going to take as long so maybe it should be a 80/20 split?

kano_updater/commands/download.py
```diff
@@ -55,10 +55,29 @@ def download(progress=None, gui=True):
...
+                90,
+                is_main=True
+            )
+        )
+        progress.start('checking')
+        check_for_updates(progress=progress)
+        if status.state == UpdaterStatus.NO_UPDATES:
+            err_msg = N_("Latest updates have been downloaded already")
+            logger.info(err_msg)
+            progress.abort(_(err_msg))
```
tombettany: Hmm, I'm not sure about this `abort()` anymore. I think that we were aborting before because this is the message that you would get if you ran the check for updates in GUI mode and there now being no updates. Admittedly, the only way that I can see of getting to this situation is if you've downloaded the updates and then for some reason we've reverted our repo to a previous version but I'm not exactly certain of the behaviour in this case.

radujipa: I've had a rethink and went through the code again. I believe it is what we'd want to do in that case - show an error message and quit the app. If you've gotten here by switching the sources, then it would effectively report there isn't an update available where you switched. The background mode would revert the state to a NO_UPDATES and the GUI mode would show no updates and revert the state again. Both would should get rid of messages probing the user to update 🤔

kano_updater/commands/download.py
radujipa: I think it was initially easier for me to write it like that and clearer to anyone looking at it for the first time. I like the compressed format more now, let me have a go

kano_updater/commands/install.py
radujipa: So this actually makes sure that was downloaded already is the latest thing by running through the `download` phase again (and consequently the `check`). So I can't know it's been downloaded before check/download-ing :)

kano_updater/commands/download.py
radujipa: Fixed and tested

kano_updater/commands/install.py
tombettany: Yeah, that's true, but we do know if it has tried to do a download before because the status will be `UPDATES_DOWNLOADED` so you could set the division based on whether it is that or it is `NO_UPDATES`.

tombettany:
Cool @radujipa, do try to add some tests and merge whenever you feel happy.

PR #271, opened 2018-01-08
Version: Bump OS version to 3.14.1

tombettany:
Bump OS version in preparation for release of v3.14.1 with resolution
policy change.

I'm guessing that we do actually want to do this but if not then somebody point it out.

Given that `rc` is the same as release right now, I'm thinking that we submit PRs for 3.14.1 to that branch and do our standard `master` rebase.


radujipa:
[As an engineer, I want to send 3.14.1 OS update](https://trello.com/c/myCSoX6E/1718-as-an-engineer-i-want-to-send-3141-os-update)

PR #273, opened 2018-02-06
Post install package cleaning

wizofe:
Created a bash script that is triggered by a hook in `apt.conf` to run post-install of `apt-get` installation procedure see: [Pull #2852](https://github.com/KanoComputing/peldins/pull/2852).


wizofe:
[(1) As a user, I want the updater to autoclean old cached files [2]](https://trello.com/c/2rgwr3GT/1744-1-as-a-user-i-want-the-updater-to-autoclean-old-cached-files-2)

kano_updater/commands/install.py
```diff
@@ -355,7 +354,7 @@ def install_standard(progress, status):
     progress.start('aux-tasks')
     run_aux_tasks(progress)
 
-
+x
```
radujipa: Woops

bin/kano-post-clean
```diff
@@ -0,0 +1,30 @@
...
+
+    # Ensure there's no lock enabled
+    sudo rm /var/cache/apt/archives/lock &&
+
+    sudo apt-get clean &&
+    sudo apt-get autoclean;
+}
+
+if [ `id -u` -ne 0 ]; then
+   echo "you need to be root"
```
radujipa: 3 space indents?

bin/kano-post-clean
```diff
@@ -0,0 +1,30 @@
...
+# Clean the retrieved .deb packages from the local repo
+# and delete all .deb files from /var/cache/apt/archives
+# to free up important space
+post_clean() {
+
+    # Ensure there's no lock enabled
+    sudo rm /var/cache/apt/archives/lock &&
+
+    sudo apt-get clean &&
+    sudo apt-get autoclean;
```
radujipa: So these are not actually triggered inside the updater? Perhaps in the `commands/install.py`?

kano_updater/commands/install.py
wizofe: too many newlines.

bin/kano-post-clean
wizofe: corrected.

bin/kano-post-clean
wizofe: nop, they are executed through the `APT-conf` instructions that the updater can run before/after `apt` runs `dpkg`. I believe it's more standard this way than add another cryptic function inside `install.py`.

bin/kano-post-clean
```diff
@@ -0,0 +1,30 @@
...
+# by enabling autocleaning
+
+
+# Clean the retrieved .deb packages from the local repo
+# and delete all .deb files from /var/cache/apt/archives
+# to free up important space
+post_clean() {
+
+    # Ensure there's no lock enabled
+    sudo rm /var/cache/apt/archives/lock &&
```
Ealdwulf: Why do you need to do that? It seems like the kind of thing that requires a very good reason.

wizofe: Because when the `DPkg::Post-Invoke` triggers, the `apt-get` manager is still running in the background (that's why it keeps a lock file in place). That means it doesn't like to run another instance of itself (the one that it will run basically the `apt-get clean`), which I am allowing by removing the lock.

The other solution is to never run this file and just do as I am describing here (inline the `apt.d`):
[Pull @2852](https://github.com/KanoComputing/peldins/pull/2852).

Ealdwulf: Okay, I don't think you should just remove the lock. So 2825 is an alternative to this PR?

wizofe: No, it goes together; My other solution was on the comments of #2852, i.e. alternatively just invoke the cleaning process through `apt.d` trigger (repasting here0:

```
DPkg::Post-Invoke { "rm -f /var/cache/apt/archives/*.deb /var/cache/apt/archives/partial/*.deb /var/cache/apt/*.bin || true"; };
APT::Update::Post-Invoke { "rm -f /var/cache/apt/archives/*.deb /var/cache/apt/archives/partial/*.deb /var/cache/apt/*.bin || true"; };

Dir::Cache::pkgcache "";
Dir::Cache::srcpkgcache
```

With this solution no other file will be added to the `/usr/bin`.

PR #274, opened 2018-02-08
Clean old core dump in kano-empty-trash

wizofe:
Implementation of a core dump cleaning function after a preset amount of days (it is set by default in 5 days, as this was the defined time-frame in the kano-feedback as well).

A unit test is under developing and going to submit as a later PR.


wizofe:
[(3) As a user, I want the updater to clean up old core dumps [2]](https://trello.com/c/EoEUrzdX/1754-3-as-a-user-i-want-the-updater-to-clean-up-old-core-dumps-2)

bin/kano-empty-trash
```diff
@@ -42,17 +43,44 @@ def empty_all_trash():
         # it thinks it must be root if sudo is in effect, but this is not the case here.
         os.system('sudo -H -u#{} SUDO_USER={} /usr/bin/kano-empty-trash '.format(uid, uname))
 
+def clean_old_dumps(dir, age):
```
skarbat: Could we change `dir` to something else? It's an internal python command

bin/kano-empty-trash
```diff
@@ -42,17 +43,44 @@ def empty_all_trash():
...
+    TODO:
+        * Get the coredump_dir directly from an environment variable or through
+        the configuration file of kernel.core_pattern 
+    '''
+    if dir:
+        for file in os.listdir(dir):
+            full_path = os.path.join(dir, file)
+            now = datetime.now()
+            mtime = datetime.fromtimestamp(os.path.getmtime(file))
+            if (file.endswith('.dump.*') or file.endswith('.dump.gz')) and ((now - mtime) > timedelta(days=age)) and os.path.isfile(full_path):
```
skarbat: Perhaps split the conditions in separate lines for readability

bin/kano-empty-trash
tombettany: 👍

bin/kano-empty-trash
```diff
@@ -42,17 +43,44 @@ def empty_all_trash():
...
+        age (int): List files older than age days old.
+    Raises:
+        OSError: Raised if the core dump dir isn there
 
+    TODO:
+        * Get the coredump_dir directly from an environment variable or through
+        the configuration file of kernel.core_pattern 
+    '''
+    if dir:
+        for file in os.listdir(dir):
```
tombettany: Likewise `file` is a builtin which should probably be avoided.

bin/kano-empty-trash
tombettany: Also, do we need both the first two clauses? Doesn't `endswith('.dump.*')` cover the `.dump.gz` case?

bin/kano-empty-trash
```diff
@@ -42,17 +43,44 @@ def empty_all_trash():
...
+            full_path = os.path.join(dir, file)
+            now = datetime.now()
```
tombettany: Do this outside the loop - it doesn't change between iterations so is unnecessary to recalculate.

bin/kano-empty-trash
```diff
@@ -42,17 +43,44 @@ def empty_all_trash():
...
+        dir (str): This is where core dump files live.
+        age (int): List files older than age days old.
+    Raises:
+        OSError: Raised if the core dump dir isn there
 
+    TODO:
+        * Get the coredump_dir directly from an environment variable or through
+        the configuration file of kernel.core_pattern 
+    '''
+    if dir:
```
tombettany: I'd swap this to be

```python
if not dir:
    raise AttributeError
```

to simplify the code and make it obvious that this is checking input arguments and all code after is the code proper.

Also, shouldn't there be a line in the comment above that it can also raise `AttributeError`?

bin/kano-empty-trash
```diff
@@ -42,17 +43,44 @@ def empty_all_trash():
...
             empty_user_trash(os.environ['HOME'])
         else:
             logger.error('HOME unset')
         sys.exit(0)
     else:
-        
         empty_all_trash()
+        try:
+            clean_old_dumps(COREDUMP_DIR, 5)
+        except AttributeError:
```
tombettany: Should we also be catching `OSError` for the case that the directory doesn't exist?

bin/kano-empty-trash
wizofe: missed that, thanks. No, the wildcard isn't supported from `endswith` unfortunately. Corrected!

bin/kano-empty-trash
wizofe: @tombettany Right! I had it outside and only changed to make the code more compact but it introduces not necessary iteration! Thanks for pointing

bin/kano-empty-trash
wizofe: Could do. The reason I haven't is that this directory is set already in `kernel.core_pattern`, so that's created by Debian by definition of the `core dump` process.

PR #277, opened 2018-02-13
Change splash image when updater is interrupted while installing.

Ealdwulf:
This PR causes the updater to switch the splash screen during the time when it is installing updates.
It puts it back if it is in any other state.
For future i18n needs, it finds the image based on the locale name.
The image is currently a placeholder.


kano_updater/paths.py
```diff
@@ -14,3 +14,5 @@
 SYSTEM_VERSION_FILE = '/etc/kanux_version'
 
 STATUS_FILE_PATH = '/var/cache/kano-updater/status.json'
+INTERRUPTED_SPLASH_PATH = '/usr/lib/python2.7/dist-packages/kano_updater/ui/media/images/'
```
tombettany: Base directory available as [kano_updater.ui.paths.IMAGE_PATH](https://github.com/KanoComputing/kano-updater/blob/master/kano_updater/ui/paths.py#L15).

kano_updater/splash.py
```diff
@@ -0,0 +1,23 @@
+#
+# splash.py
+#
+# Copyright (C) 2015 Kano Computing Ltd.
```
tombettany: 2018

kano_updater/status.py
```diff
@@ -12,7 +12,7 @@
 from kano.logging import logger
 
 from kano_updater.paths import STATUS_FILE_PATH
-
+from kano_updater.splash import set_splash_interrupted, clear_splash
```
tombettany: Bump the copyright of this file.

kano_updater/paths.py
tombettany: Also bump copyright.

debian/control
```diff
@@ -25,7 +25,7 @@ Depends:
     python-apt,
     schedtool,
     python-pip,
-    kano-i18n (>= 3.5-1),
+    kano-i18n (>= 3.15-1),
```
radujipa: *3.15.0

kano_updater/paths.py
Ealdwulf: I wondered why it wasn't there already. Okay, I've eliminated use of this file, so no need to bump copyright.

kano_updater/status.py
Ealdwulf: Done.

debian/control
Ealdwulf: Yes - or at least, corresponding to kano-i18n.

PR #279, opened 2018-02-14
Stop the boot loading indicator if resuming update installation

radujipa:
After a power failure during an install, the subsequent boot would
launch the Updater but the loading indicator would be over it for
a while until it terminates by itself.

Same case as: https://github.com/KanoComputing/os-dashboard/blob/master/bin/kano-dashboard-supervisor#L288


bin/kano-updater
```diff
@@ -186,6 +186,7 @@ def main():
     if gui_mode:
         # Hide the home button only if we are on Desktop mode
         run_bg('systemctl --user status kano-desktop && /usr/bin/kano-home-button-visible no')
+        run_bg('kano-stop-splash dashboard')
```
skarbat: Yeah, looks okay, but I think we'll need a few more things than that.

 * Use the new better "sudo systemctl stop kano-boot-splash" - or new equivalent name
 * Move the systemd service that stops the splash to a later stage, as I noticed weird UI behaviour on recovery
 * Probably disable plymouth, which seems to try to do things on recovery

radujipa:
Found a better fix https://github.com/KanoComputing/kano-desktop/pull/290

PR #280, opened 2018-02-14
Avoid running the CKC power check when coming from boot

radujipa:
In case of a power failure during an update, the subsequent boot
will relaunch the Updater very early. As such, on a CKC the popup
asking users if the kit was plugged in would potentially trigger.
This adds an option to avoid running that check at that point.


radujipa:

PR #275, opened 2018-02-09
Resume interrupted updates at boot

radujipa:
This PR was to ensure that the `kano-updater ui boot-window` was up to date and capable of continuing an interrupted install. Some indirect improvements where made along with that. Changes are organised in commits to help with reviewing.

### Removed install confirm window at bootup

This simply removes the user confirmation window when updates have
been scheduled. Previously at shutdown, if there was an update
available the user would be asked if they wanted to install it
then or later. Subsequently, they would get asked again at bootup.
Now, the Dashboard will constantly show a message about installing
and this popup is no longer useful.

### Remove broken and unused post-update bootup changes dialog

It seems like there was an old mechanism of showing what was
installed after an update with an HTML page and a WebApp dialog.
When this was manually triggered, the Updater exited with a
traceback. Removing this part of the code along with any other

### Improve app launch time by eliminating D-Bus iface retries

This sets the retry_count=0 in other methods that were missed in https://github.com/KanoComputing/kano-updater/commit/10c1b9923a6d08853e36dac14fbcbdd699c79d29

### Fix binary executable permissions

For some reason these two binaries did not have executable
permissions??? **How did this ever work?**

### Others

 - Removed unused app assets (pngs, gifs)
 - Code styling and removing imports of unused modules

PR #281, opened 2018-02-15
Fixes an issue with not launching background updates

radujipa:
I initially got my ifs wrong and didn't start the download if the
system already booted for over 3 minutes.

Related to: https://github.com/KanoComputing/kano-updater/pull/275


radujipa:

PR #282, opened 2018-02-15
Added post-clean script for apt upgrade

wizofe:
Implement a cache cleaning functionality after the installation and upgrade of `apt` via/and `dpkg`. 

It is using the `apt` `daemon` post upgrade and install process to be sure that it's called either through the `kano-updater` or via any manual package install/upgrade from the user.

INFO: Changed per @skarbat suggestion to include a diagnostic/informative message before igniting the process.


wizofe:
[(2) As a user, I want the updater to autoclean old cached files [2]](https://trello.com/c/2rgwr3GT/1744-2-as-a-user-i-want-the-updater-to-autoclean-old-cached-files-2)

conf/apt.conf.d/99clean
```diff
@@ -0,0 +1,7 @@
+# Keep the post installation environment clean
+# and completely empty the cache after upgrade/update
+# for more options see: man (5) apt.conf 
+
+DPkg::Post-Invoke { "echo Kano cache cleanup...;rm -f /var/cache/apt/archives/*.deb /var/cache/apt/archives/partial/*.deb /var/cache/apt/*.bin || true"; };
+APT::Update::Post-Invoke { "echo Kano post cleanup...;rm -f /var/cache/apt/archives/*.deb /var/cache/apt/archives/partial/*.deb /var/cache/apt/*.bin || true"; };
```
radujipa: Don't these `echo`s need the string to be quoted? Also, I'd perhaps use the `logger` for this to be in the journal.

conf/apt.conf.d/99clean
```diff
@@ -0,0 +1,7 @@
+# Keep the post installation environment clean
+# and completely empty the cache after upgrade/update
+# for more options see: man (5) apt.conf 
+
+DPkg::Post-Invoke { "echo Kano cache cleanup...;rm -f /var/cache/apt/archives/*.deb /var/cache/apt/archives/partial/*.deb /var/cache/apt/*.bin || true"; };
```
skarbat: ouch... sorry. single quotation marks for the echo literals?

conf/apt.conf.d/99clean
skarbat: Personally I think calling logger in the middle of an apt hook can be dangerous.

wizofe: Thinking if you agree, to merge it as it is, with a simple echo to show info and only change in future if there is a need.

PR #284, opened 2018-02-15
Moving kano-os-recovery as a dependency instead

skarbat:
Instead of installing the recovery package by a command, do it through APT dependencies.


skarbat:
[(4) As a user who unplugs during an upgrade, I don't want to ever see the black screen again [5]](https://trello.com/c/tDXcPWkU/1767-4-as-a-user-who-unplugs-during-an-upgrade-i-dont-want-to-ever-see-the-black-screen-again-5)

PR #286, opened 2018-02-16
Disk Space: Revert packages in reverse order

tombettany:
When checking for disk space requirements, every package is marked for
update and then reverted to its previous state after the space
requirement is calculated. Prevent reversion errors by restoring their
original state in the opposite order to that which set it in the first
place. Also, don't bother restoring package states which were already
marked for update before the whole process began.


kano_updater/apt_wrapper.py
```diff
@@ -259,13 +259,15 @@ def get_required_upgrade_space(self, priority=Priority.NONE):
...
+            if state != AptPkgState.MARKED_UPGRADE:
+                orig_state.append((pkg, state))
+                pkg.mark_upgrade()
 
         space = self._cache.required_download + self._cache.required_space
 
-        for pkg, state in orig_state:
+        # Restore package states in reverse order
+        for pkg, state in reversed(orig_state):
             AptPkgState.restore_pkg_state(pkg, state)
```
radujipa: Do you want to put this `restore` under a similar `if` as above?

tombettany: Don't need to because they would have only been appended to `orig_state` if they weren't already marked for update.

PR #287, opened 2018-02-19
Daily apt cache cleanup with systemd

radujipa:
Firstly, this creates a daily systemd timer which triggers the
daily-cleanup target, then services can be installed on that target.
Secondly, this creates and enables an apt-cleanup service which
runs autoremove and autoclean.


debian/kano-updater.postinst
```diff
@@ -46,6 +46,9 @@ case "$1" in
             grep -m 1 '^deb http://.*.kano.me/archive/' $KANO_SOURCES | \
                 sed -r "s|^deb http://([a-z]*).kano.me.*$|deb http://\1.kano.me/archive/ release-urgent main|" >> $KANO_SOURCES
 
+        # Enable systemd services and timers.
+        systemctl enable apt-autoclean-autoremove.service
+        systemctl enable daily-cleanup.timer
```
wizofe: Also `&& systemctl start apt-autoclean-autoremove.service`? Same for timer

conf/systemd/system/daily-cleanup.timer
```diff
@@ -0,0 +1,18 @@
...
+#
+# A daily occuring event to trigger cleanup routines.
+
+
+[Unit]
+Description=Daily Cleanup Timer
+
+[Timer]
+Unit=daily-cleanup.target
+OnCalendar=*-*-* 12:00:00
```
wizofe: Would you like to delay this after boot something like `OnBootSec=10m` as not to clash with the `updater` resuming (as described in `70-kano-updater-download` hook)?

debian/kano-updater.postinst
radujipa: For most, this will be installed via the Updater in an update. Starting the them would actually trigger and execution when there's going to be a reboot immediately after anyway. I don't think it's needed here.

conf/systemd/system/daily-cleanup.timer
radujipa: Definitely, good catch 👍

wizofe:
I would change the `daily-cleanup.timer` to `apt-cleanup.timer` and `apt-autoclean-autoremove` to `apt-cleanup.timer` to keep it short and uniform but I don't know if you prefer it as it is, or if there's a specific naming idea you are currently following with `systemd` and I am unaware.

wizofe:
Looks good to me! Only thing, I don't know how it will behave in a situation that `apt` locks while the process is needed by the `updater`.

radujipa:
@wizofe I used that naming for the `target` (the `timer` should have the same name, different suffix) because it's more generic that way. We could hook up `kano-empty-trash` or whatever to it and it would still make sense. It's an event to cleanup any system stuff.

radujipa:
A safer option was implemented here: https://github.com/KanoComputing/kano-updater/pull/290

PR #289, opened 2018-02-20
Revert wizofe:wip/apt-cache-clean

radujipa:
Reverts https://github.com/KanoComputing/kano-updater/pull/282
addressed in https://github.com/KanoComputing/kano-updater/tree/apt-hook-autoclean


debian/changelog
```diff
@@ -11,6 +9,7 @@ kano-updater (3.15.0-0) unstable; urgency=low
   * Moved background update downloads into project through dhcpcd hook
   * Switch the boot splash to inform the user when the update was interrupted
   * Intelligently gather disk space required for installation
+  * Added functionality to empty-trash bin as to clean old core dumps
```
tombettany: We need this line?

radujipa: Yup. I just moved that down, it was added somewhere else.

PR #292, opened 2018-02-23
Configure autologin during update installation

radujipa:
When the Updater is relauched at bootup to potentially continue an
interrupted update on a multi-user system, the Greeter will come
up and prevent the update from continuing until the a user logs in.
Perform an autologin on the user account that started the update
to continue the installation process.

### Added tests for checking LightDM autologin policy function calls

This essentially only checks if the autologin functions are being
called as expected from the Updater during an install. They don't
test anything regarding the before/after state, which means:

 1. kano_init library tests the functions fully to make sure they
    work as intended.
 2. Integration between kano_updater and kano_init is not handled
    and not implemented in the Updater unit tests. If this were
    handled here, then changes in kano_init that would break the
    Updater would only be visible when rebuilding the Updater.

Requires: 
https://github.com/KanoComputing/os-dashboard/pull/372
https://github.com/KanoComputing/kano-settings/pull/499


radujipa:

radujipa:
This PR is now ready 🎉

kano_updater/status.py
```diff
@@ -125,12 +127,20 @@ def save(self):
...
+        # When installing updates, configure the recovery stategy for the next
+        # boot in case of power failure. This sets a different splash screen
+        # during bootup and configures the system to autologin as the user in
+        # order to start the Updater immediately. See kano-ui-autostart.
+        if self._state in [self.INSTALLING_UPDATES, self.INSTALLING_INDEPENDENT]:
+            logger.debug('Configuring recovery stategy for next boot')
             set_splash_interrupted()
+            enable_console_autologin(get_user_unsudoed())
+            set_ldm_autologin(get_user_unsudoed())
+            enable_ldm_autostart()
```
tombettany: Can we add a new function to `kano_init` for `set_user_autologin()` (or something like that) which does these three functions in order; the updater knows too much about the internals of the login process here: these shouldn't be part of the public API.

kano_updater/status.py
```diff
@@ -125,12 +127,20 @@ def save(self):
...
-        if self._state in [self.INSTALLING_UPDATES,
-                           self.INSTALLING_INDEPENDENT]:
+        # When installing updates, configure the recovery stategy for the next
+        # boot in case of power failure. This sets a different splash screen
+        # during bootup and configures the system to autologin as the user in
+        # order to start the Updater immediately. See kano-ui-autostart.
+        if self._state in [self.INSTALLING_UPDATES, self.INSTALLING_INDEPENDENT]:
+            logger.debug('Configuring recovery stategy for next boot')
             set_splash_interrupted()
+            enable_console_autologin(get_user_unsudoed())
```
Ealdwulf: So, the only concern I have here is whether any of these can fail. 
In particular `get_user_unsudoed()` might not always work.
The typical prescence of `/root/.kano-logs/kano-updater.log`  suggests that sometimes the updater is run in a context in which it can't figure out what the user is. Given that the state is modified in every context in which the updater runs, maybe this should be run in a try-except block.

Also - and this is my fault for originally putting the splash here - it' is a little obscure to have the state making these system changes. Perhaps it would be clearer if all of these calls were separated out into  functions like `updater_system_prepare` `updater_system_release` (or something like that) so it's more obvious.

radujipa:
@Ealdwulf @tombettany good points, thank you. I've separated out the recovery setup into another module.

PR #293, opened 2018-02-26
Recovery flow for interrupted updates

tombettany:
Only half implemented - feel free to modify as you wish.

First iteration of the recovery flow script. This implements the basic
recovery flow detailed in the [updater recovery specification](https://docs.google.com/presentation/d/13IwXal8iJYRWBXrCoCI-dD7DTF2WkRxZIdiYxSUroQY/edit?ts=5a8e951c#slide=id.g310fc115b5_0_82)

Outstanding points are:
 - `prepare_for_hard_shutdown()` once an error splash was shown
 - Animated or Static splash screens for Recovery, Updater, Successful
 - Logging
 - Testing

### Requires:
https://github.com/KanoComputing/kano-desktop/pull/292


radujipa:
[As a user, I want the update recovery to give various error codes and messages at different stages](https://trello.com/c/qTmfIGmX/1784-as-a-user-i-want-the-update-recovery-to-give-various-error-codes-and-messages-at-different-stages)

bin/kano-updater-recovery
```diff
@@ -0,0 +1,284 @@
...
+    from kano_updater.status import UpdaterStatus
+    from kano_updater.utils import remove_pid_file
+
+    old_status = clean(dry_run=True)
+    status = UpdaterStatus.get_instance()
+
+    if old_status == UpdaterStatus.INSTALLING_UPDATES:
+        status.notifications_muted = True
+        status.state = UpdaterStatus.UPDATES_AVAILABLE
+        status.save()
```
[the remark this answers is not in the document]
radujipa: You're totally right, the pidfile stuff does not apply here - I've removed it. Will look at putting a lock on the Updater status file.

radujipa:
With the exception of `prepare_for_hard_shutdown()` which I'm looking at, this PR is now reviewable - @tombettany @wizofe @skarbat @Ealdwulf

radujipa:
[(3) As a user, I would like the update recovery to inform me of network connectivity issues](https://trello.com/c/ZVPcVT5N/1787-3-as-a-user-i-would-like-the-update-recovery-to-inform-me-of-network-connectivity-issues)

bin/kano-updater
```diff
@@ -115,8 +115,9 @@ def schedule_install(gui=False, confirm=True, splash_pid=None):
         confirm = False
 
         title = "Updater"
-        description = "Urgent updates have been downloaded! We'll need to install these as soon" \
-                      " as possible to ensure the best experience."
+        description = "Urgent updates have been downloaded! We'll need to" \
```
skarbat: do we want to `i18n` this string?

bin/kano-updater
```diff
@@ -330,8 +314,8 @@ def main():
                     # Show dialogue
                     kdialog = KanoDialog(
                         "Updater",
-                        "My brain already has the latest bits and bytes - high" \
-                        " five!",
+                        "My brain already has the latest bits and bytes"
```
skarbat: likewise i18n, if applies

bin/kano-updater-recovery
```diff
@@ -0,0 +1,351 @@
...
+    """
+    Shows the splash for a successful recovery
+
+    Returns:
+        int: Return code. Should be one of the members of the :class:`RC` class
+    """
+
+    _start_systemd_service('recovery-success-splash')
+    import time
+    time.sleep(0.2)
```
skarbat: would it be more practical to move these `0.2` as a variable above, along with MINUTE and HOUR?

bin/kano-updater-recovery
```diff
@@ -0,0 +1,351 @@
...
+    Assumes the state was already checked to be INSTALLING_UPDATES.
+
+    Returns:
+        subprocess.Popen: Launched updater process
+    """
+
+    import os
+    import subprocess
+
+    return subprocess.Popen(
```
skarbat: Can we add a short comment on why we need to `setsid` the new process?

bin/kano-updater-recovery
```diff
@@ -0,0 +1,351 @@
...
+
+
+def main():
+    """Main recovery program."""
+
+    from kano_updater.status import UpdaterStatus
+    if not UpdaterStatus.get_instance().is_recovery_needed():
+        return RC.SUCCESS
+
+    # Check if any UFx is being shown, e.g. kano-os-recovery had a timeout.
```
skarbat: I am guessing we are preparing for a shutdown initiated by an external process? Can we briefly explain this?

conf/systemd/system/recovery-boot-splash.service
```diff
@@ -0,0 +1,22 @@
+# recovery-boot-splash.service
+#
+# Copyright (C) 2018 Kano Computing Ltd.
+# License: http://www.gnu.org/licenses/gpl-2.0.txt GNU GPL v2
+#
+# Runs initial the splash screen for the recovery mechanism when the system is
```
skarbat: swapped words - `the initial`

debian/kano-updater.install
```diff
@@ -14,3 +14,4 @@ locale/* /usr/share/locale
 
 conf/apt etc/
 conf/dhcpcd-hooks lib/dhcpcd/
+conf/systemd usr/lib/
```
skarbat: should this be `conf/ usr/lib`, in order to land at `/usr/lib/systemd/system`?

debian/kano-updater.postinst
```diff
@@ -17,6 +17,7 @@ case "$1" in
 
         # Create custom sudoers file
         echo "%sudo   ALL=(root) NOPASSWD: /usr/bin/kano-updater" > $TMP_FILE
+        echo "%sudo   ALL=(root) NOPASSWD: /usr/bin/kano-updater-recovery" >> $TMP_FILE
```
tombettany: Can we move away from this `TMP_FILE` mess and instead create a `kano-updater.conf` file in the repo and install it with the `kano-updater.install` file. `dh_fixperms` should already be handling [permissions](https://github.com/Debian/debhelper/blob/master/dh_fixperms#L158) but at the very most, the contents of this `postinst` should set permissions.

tombettany: Also changing this file should trigger a copyright bump.

kano_updater/return_codes.py
```diff
@@ -0,0 +1,89 @@
...
+    NOT_PLUGGED_IN = 24
+
+    NO_NETWORK = 30
+    CANNOT_REACH_KANO = 31
+    HANGED_INDEFINITELY = 32  # TODO
+    SIG_TERM = 33
+    NOT_ENOUGH_SPACE = 34
+
+
+class RcState(object):
```
tombettany: This just looks a bit weird because it isn't `RCState`.

bin/kano-updater-recovery
```diff
@@ -0,0 +1,351 @@
...
+            rc = show_err(RC.UF6)
+            break
+
+        time.sleep(1)
+
+    # If the script has gotten to this point, a UFx is being shown. It's
+    # expected that the recovery shows the error indefinitely until the user
+    # unplugs to power off.
+    prepare_for_hard_shutdown()
+    infinite_loop()
```
tombettany: If we've popped up one of the error codes using the `recovery-err-splash@.service` unit, the message won't timeout so why do we need to keep this alive with the infinite loop?

radujipa: This will be because this script is started from `kano-ui-autostart`. As such, if the script terminates, the other will continue to go to Dashboard. Maybe when we convert that one into a `.target` and move everything into `.services` thereafter..

debian/kano-updater.install
radujipa: The above actually says that the `conf/systemd` folder (and everything in it) will be installed at `/usr/lib/`. Each folder in `conf/` has a slightly different install path.

bin/kano-updater
radujipa: We should probably do another pass of the i18n state of the Updater. Will do that in another PR

bin/kano-updater-recovery
radujipa: So when there is an error during the entire flow (split between `kano-os-recovery` and `kano-updater-recovery`) the spec is to show a `recovery-err-splash` indefinitely. This means that the user is expected to unplug and (not) try again depending on the error code shown.

Here, we check whether `kano-os-recovery` has failed to quit `kano-updater-recovery` early. Updated the comment.

bin/kano-updater-recovery
radujipa: This was to assign an id to the entire process group to make stopping it more reliable. Updated the comment.

conf/systemd/system/recovery-boot-splash.service
radujipa: Thanks, fixed :)

debian/kano-updater.postinst
radujipa: I like the idea of the sudoers conf file. Will tackle that separately

Thanks, fixed the copyright here

kano_updater/return_codes.py
radujipa: Initially that looked weird but not so much now. Changed it everywhere

debian/kano-updater.install
skarbat: You're right, sorry

skarbat:
👍

PR #294, opened 2018-02-27
Remove initial docopt install

Ealdwulf:
This PR removes the initial attempt to install docopt. All version we can now update from already have docopt.


Ealdwulf:
Done. Sadly, adding the changlog means this PR is no longer entirely composed of removed lines.

tombettany:
I'm a little late to the game here but don't we want to add a Debian dependency on the Debian package of `docopt` rather than just relying on it being present on their systems?

radujipa:
Yup, we'll do that once all pip pkgs will be converted to debs.

tombettany:
👍

PR #296, opened 2018-02-28
pip-transition.md

Ealdwulf:
This PR adds a doc about the pip->deb transition. However the main point of this PR is for people to review whether the transition steps make sense, are there other things which need checking, etc.


docs/pip_transition.md
```diff
@@ -0,0 +1,76 @@
...
+  
+Steps
+
+### 1.  Build python packages and add them to dr
+* Building:
+   * py2deb seems to work reasonably well. The only caveat is that it needs `pip-accel` to be manually installed first `pip install pip-accel` as otherwise you get a version of pip which pip-accel is not compatible with. Then do `pip install py2deb`.
+* Choice of packages and versions:
+  * It is not necessary to install pip once we transition.
+  * Some packages also already have raspbian versions. I propose to build new ones where the pip ones we currently install are newer, but not where they are older. There are only two where the raspbian version is newer (pam and futures). If this proves to be a problem, we can build the exact version and pin in the debian control file.
+
```
Ealdwulf: On reflection, I think we should just pin all the new deb versions to exactly the same as the current pip versions. We can always unpin them later, and there's going to be enough to QA in this release.

docs/pip_transition.md
```diff
@@ -0,0 +1,76 @@
+## Transition from using pip packages
+
+This documents the proposed transition to no longer using pip packages (fully debianised system)
+
+### Rationale
```
radujipa: I guess another one here is that some packages are installed via both deb and pip which causes `dpkg --verify` to freak out. You can see this here https://github.com/KanoComputing/kano-settings/pull/496/files

docs/pip_transition.md
```diff
@@ -0,0 +1,76 @@
...
+
+
+### 2. Transition from pip to deb packages.
+
+This must:
+
+  * Install the new deb packages
+  * Ensure the old pip packages don't override them.
+  
+It turns out by deafult pip packages do override deb packages. Python load modules from `sys.path`
```
radujipa: *default

docs/pip_transition.md
```diff
@@ -0,0 +1,76 @@
...
+   `/usr/lib/python2.7/dist-packages/kano-pip-compat.pth`
+ containing `/usr/local/lib/python2.7/dist-packages.pip-fallback`
+This adds that path to sys.path after `/usr/lib/python2.7/dist-packages`.
+
+##### 3 Pre-update scenario Move old pip packages to fallback path 
+```
+mkdir -p /usr/local/lib/python2.7/dist-packages.pip-fallback
+mv /usr/local/lib/python2.7/dist-packages{,.pip-fallback}
+```
+##### 5 deb packages installed as part of main install step
```
radujipa: 4?

docs/pip_transition.md
```diff
@@ -0,0 +1,76 @@
...
+ * Stop shipping our built python deb files once newer ones become available in raspbian\
+ 
+ 
+### 4 Things that need to be checked:
+
+  * No other calls to pip in the system outside the updater
+  * No other useful files in `/usr/local/lib/python2.7/dist-packages`
+  * newer versions of pam and futures don't cause problems
+  * All versions correct on installed system
+
```
radujipa: So a few things that need to happen to the Updater and image build for this:

 - Remove `kano-install-pip-deps` from peldins/stage-kano-jail 
 - Remove `kano-install-pip-deps` from kano-updater/bin
 - Remove `python_modules` from kano-updater
 - Remove everything `pip` related from kano-updater sources: download, install, paths, etc
 - Add all py-debs as deps to a subpackage of the Updater for now and slowly transition them out

docs/pip_transition.md
Ealdwulf: Yeah. To note our verbal discussion: That was because `pip uninstall` will happily remove any pip packages, not just ones it installed itself. This is why I propose manually moving `/usr/local/lib/python2.7/dist-packages`: I couldn't convince myself that we could use `pip uninstall` reliably to uninstall the existing pip packages without breaking stuff.

wizofe: In fact, even though not confirmed, my suspicion was that `pip uninstall` was one of the reasons of the crash; when I stopped the installer between the uninstall phase it didn't properly handled the `dist-packages` locations. That shows it doesn't work as expected especially concerning handling existing `pip` depedencies.

Ealdwulf:
Thanks for the review @radujipa . I've pushed some commits which:
 * Fix typos
 * Change to specifying that the deb versions should be pinned
 * change to leave an empty directory `/usr/local/lib/python2.7/dist-packages`

docs/pip_transition.md
```diff
@@ -0,0 +1,79 @@
...
+```
+##### 4 deb packages installed as part of main install step
+A new package `kano-os-pip-transition-depends` which depends on the new deb packages will be added.
+
+The code in the updater which downloads/installs pip packages needs to be removed before this update.
+
+### 3. Pay down tech debt:
+ * Add correct dependencies to other packages.
+ * Stop shipping our built python deb files once newer ones become available in raspbian\
+
```
wizofe: I believe that it would be easier for us to always ship our own python `deb`'s as that way we can always ensure that we have tested the required versions and are not dependent on the Raspbian releases.

Ealdwulf: That might be true. We should definitely build them for now, anyway, so that's a decision for later.

docs/pip_transition.md
Ealdwulf: I agree with all this. Rather than adding it to the doc, I have copied it into the trello ticket, along with all of the other concrete stuff in this doc.

wizofe:

PR #297, opened 2018-03-05
WIP: Remove pip from kano-updater

wizofe:
For more details see the flow on the following trello attachment and #296 
- Add `docopt` depedency through `debian/control`
- Remove `python_modules`
- Remove `kano_pip_install_deps`
- Correct estimation of `free_space.py` without the `pip` packages
- Remove test fixtures


wizofe:

debian/control
```diff
@@ -23,6 +23,7 @@ Depends:
     gir1.2-gtk-3.0 (>= 3.10.2),
     kano-settings (>= 1.3-1),
     python-apt,
+    python-docopt,
```
radujipa: `python-pip` will also be in here presumably

debian/kano-updater.postinst
```diff
@@ -34,9 +32,7 @@ case "$1" in
             rm $OLD_UPDATER_STATUS
         fi
 
-        pip install --upgrade $PIP_DEPENDENCIES
-
-        # Add the urgent repo to sources - if not already existent (jessie users)
+                # Add the urgent repo to sources - if not already existent (jessie users)
```
radujipa: Indenting?

bin/kano-install-pip-deps
```diff
@@ -1,70 +0,0 @@
-#!/usr/bin/env python
```
tombettany: Where were we even using this script?

wizofe: It was used by `peldins/stage-kano-jail.sh`, that's all I know :)

skarbat: Looks like it: https://github.com/search?q=kano-install-pip-deps&ref=simplesearch&type=Code&utf8=%E2%9C%93

I'll fix the QA test

tombettany: Ah, gotcha. Glad to see it go then :)

debian/kano-updater.postinst
wizofe: Merci, corrected.

conf/dist-packages/kano-pip-compat.pth
```diff
@@ -0,0 +1,7 @@
+# Copyright (C) 2018 Kano Computing Ltd.
+# Licensed under the GNU GPL v2
```
radujipa: Header format is usually:

```
# <filename>
#
# Copyright (C) 2018 Kano Computing Ltd.
# License: http://www.gnu.org/licenses/gpl-2.0.txt GNU GPL v2
```

:)

kano_updater/utils.py
```diff
@@ -135,21 +128,37 @@ def run_for_every_user(cmd):
         run_cmd_log("sudo su -c '{cmd}' - {user}".format(cmd=cmd, user=user))
 
 
-def is_server_available():
-    install_ping()
-    import ping
-
-    lost_packet_percent = ping.quiet_ping(REPO_SERVER, timeout=3, count=5)[0]
+def is_server_available(REPO_SERVER, count = 5):
```
radujipa: I would honestly suggest not reimplementing this function with `subprocess` as opposed to using `ping`. If it can be done through the module I think that'll be more maintainable and safer. In that case you'd also have to make the `kano-updater` package depend on `python-ping` in the control file.

tombettany: +1

conf/dist-packages/kano-pip-compat.pth
```diff
@@ -0,0 +1,7 @@
...
+#
+# Function: Adds the old pip file depedencies in the following path
+# to be used in case of failure. See scenarios.py: 3.14.1 to 3.15
+
+/usr/local/lib/python2.7/dist-packages.pip-fallback
```
radujipa: Is this file being installed in the `kano-updater.install` ?

debian/control
```diff
@@ -39,3 +39,19 @@ Package: kano-updater-i18n-orig
 Architecture: all
 Description: Data for working on translations of kano-updater
 Multi-Arch: foreign
+
+Package: kano-os-pip-transition
```
tombettany: Not too keen on this package name - I know that right now we are doing it to transition from pip to Debian packages but really it is just a metapackage of pip modules which are required to be installed. Do we even want this package - should these dependencies not just go into the `kano-os` metapackage?

radujipa: The idea of the subpackage is to temporarily help transition `python_modules` with a debian way. Ideally, all of the depends of this one would be transitioned to other packages that actually need them.

wizofe: That's it, mostly a temporary package, to safeguard that everything works as planned and is easier to debug. For the next release, it shouldn't be there; moreover attach the dependencies directly on the updater control.

I could add a comment there, if there was a way to add comments in a `debian/control` file 🤔

kano_updater/scenarios.py
```diff
@@ -965,4 +977,10 @@ def beta_3_14_0_to_beta_3_14_1(self):
         pass
 
     def beta_3_14_1_to_beta_3_15_0(self):
-        pass
+        try:
+            import shutil
+            import os
+
+            shutil.copy(PIP_CONF, PYDIST_PATH)
```
Ealdwulf: Why are we installing the .pth file in a post-update? Doesn't this mean there is a gap when those modules are not available? What happens if the user pulsl theplug just after the pre-update?

debian/changelog
```diff
@@ -16,6 +16,7 @@ kano-updater (3.15.0-0) unstable; urgency=low
   * Implemented a Recovery Mode for interrupted updates
   * Removed deprecated code and general cleanup
   * Implemented various exit codes for the Updater
+  * Removed pip and replaced with debian package installation on depedencies
```
Ealdwulf: s/depedencies/dependencies

kano_updater/scenarios.py
tombettany: Is `PYDIST_PATH` defined anywhere?

conf/dist-packages/kano-pip-compat.pth
radujipa: bump

debian/control
```diff
@@ -17,14 +17,15 @@ Package: kano-updater
...
 Depends:
     ${misc:Depends},
+    kano-os-pip-transition,
     kano-toolset (>= 3.7.0-1),
     python,
     libkdesk-dev,
     gir1.2-gtk-3.0 (>= 3.10.2),
     kano-settings (>= 1.3-1),
     python-apt,
+    python-docopt (= 0.6.2),
```
radujipa: `python-ping` should also be up here

kano_updater/utils.py
```diff
@@ -134,23 +127,19 @@ def run_for_every_user(cmd):
...
+    """ Pings REPO_SERVER to diagnose packet loss
+    and server availability.
 
+    Args:
+        REPO_SERVER: is the pre-set address
+        count: Number of times to ping the server
 
-def install_ping():
-    try:
-        import ping
```
Ealdwulf: Unless I'm missing something, ping doesn't seem to be imported now?

kano_updater/scenarios.py
wizofe: Thanks, it's defined in `paths` as `PYLIBS_DIR`, corrected.

kano_updater/utils.py
wizofe: Thanks, chopped off too much.

kano_updater/scenarios.py
wizofe: The problem I have here is that `pre-update` is moving the `dist-packages` to the fallback directory and as a result the `pth` file that is installed through the debian `control` file. So, I decided to install the `pth` in a system directory and copy it `post-update`.

Do you have a better idea/recommendation for this?

Ealdwulf: You can put the `.pth` file in `/usr/lib/python2.7/dist-packages` instead of in `/usr/local/lib`

wizofe: Thanks for the recommendation @Ealdwulf. Done and squashed.

PR #298, opened 2018-03-12
Add monitor

Ealdwulf:
This PR:
 - wraps kano-updater in a monitoring program
 - changes some `from` imports to simple imports to avoid circular dep problems.

Note that it moves `bin/kano-updater` to `bin/kano-updater-internal`.
For ease of review, this is done in a separate commit, so that examining the second commit will show the substantive changes.

Some testing still to be done.


kano_updater/monitor.py
```diff
@@ -0,0 +1,165 @@
+# monitor.py
+#
+# Copyright (C) 2014-2018 Kano Computing Ltd.
```
radujipa: *2018

kano_updater/monitor.py
```diff
@@ -0,0 +1,165 @@
...
+import signal
+import os
+from kano.logging import logger
+import kano_updater.utils
+
+MONITOR_TIMEOUT = 20 * 60
+TIMEOUT_RC = 105
+
+
+class monitorPids:
```
radujipa: I'd suggest capitalising the class name, `MonitorPids`

kano_updater/monitor.py
```diff
@@ -0,0 +1,165 @@
...
+                curr = pids.copy()
+                for p in pids:
+                    curr = curr.union(closure(coll[p]))
+                return curr
+            return closure(set([self.top_pid]))
+
+        except subprocess.CalledProcessError:
+            return set()
+
+    def isChanged(self):
```
radujipa: *is_changed

kano_updater/monitor.py
```diff
@@ -0,0 +1,165 @@
...
+    if not monitor(subproc, MONITOR_TIMEOUT):
+        return subproc.returncode
+    else:
+        if '--gui' in cmdargs:
+            from kano.gtk3 import kano_dialog
+            kdialog = kano_dialog.KanoDialog(
+                _("Update error"),
+                _("The updater seems to have got stuck. Press OK to reboot"))
+            kdialog.run()
+            os.system('systemctl reboot')
```
radujipa: I expect this to be more for the autoupgrades, perhaps add a `return RC` here

kano_updater/utils.py
```diff
@@ -585,3 +585,21 @@ def verify_kit_is_plugged():
             ).run()
 
     return is_plugged and not is_battery_low
+
+
+class signalPoll:
```
radujipa: As above, I suggest capitalising the class name `SignalPoll`

kano_updater/monitor.py
skarbat: Suggest we add something like this:

 * https://github.com/KanoComputing/kano-updater/blob/master/kano_updater/ui/views/finish.py#L71-L75

In order to play well in the emulated environment on the autoupgrades.
I am trying to match how will all this fit in that sandbox...

kano_updater/monitor.py
```diff
@@ -0,0 +1,165 @@
...
+
+        else:
+            return TIMEOUT_RC
```
radujipa: Shouldn't this be `RC.HANGED_INDEFINITELY` from `kano_updater.return_codes`?

radujipa:
Also, could you add a note in the changelog about this feature please?

kano_updater/monitor.py
```diff
@@ -0,0 +1,165 @@
...
+        try:
+            stdout = subprocess.check_output(["/bin/ps", "-eo", "ppid,pid"])
+            coll = defaultdict(set)
+
+            for line in stdout.split('\n'):
+                items = line.split()
+                if len(items) >= 2:
+                    ppid, pid = items[:2]
+                    if ppid == 'PPID':
+                        continue
```
tombettany: Huh? You don't seem to be doing anything with `ppid` or `pid` - the `try` block below is outside the loop so it will only be done for the last PID.

kano_updater/monitor.py
```diff
@@ -0,0 +1,165 @@
...
+            try:
+                coll[int(ppid)].add(int(pid))
+            except:
+                pass  # ignore int conversion failure
+
+            def closure(pids):
```
tombettany: While this may be a closure, that name isn't particularly helpful to understand what is going on - maybe call it something like `get_child_pids()` or something (although that might conflict with the main function name. Also a docstring would be helpful to understand the arguments required.

kano_updater/monitor.py
```diff
@@ -0,0 +1,165 @@
...
+
+def monitor(watchproc, timeout):
+    """
+    Monitor a process tree. If the (recursive) set of child processes changes
+    or we are sent a SIGUSR1, we note that it is making progress.
+    if it has not made progress for `timeout` seconds, or the process
+    finished, exit.
+
+    Returns true if we timed out and false if the process finished
+
```
tombettany: Arguments in either [this form](http://dev.kano.me/docs/kano-toolset/_modules/kano/utils/file_operations.html#touch) so sphinx can parse them.

kano_updater/monitor.py
```diff
@@ -0,0 +1,165 @@
...
+                _("Update error"),
+                _("The updater seems to have got stuck. Press OK to reboot"))
+            kdialog.run()
+            os.system('systemctl reboot')
+
+        else:
+            return TIMEOUT_RC
+
+
+def manual_test_main(argv):
```
tombettany: Maybe instead add a directory into the `tests` directory for manual tests and put this there?

kano_updater/utils.py
```diff
@@ -585,3 +585,21 @@ def verify_kit_is_plugged():
...
+    # A class to allow using signals without globals
+
+    def __init__(self, sigNum):
+        self.sigNum = sigNum
```
tombettany: The Pythonic variable should really be called `sig_num`.

kano_updater/utils.py
tombettany: +1 to above and do you want to split this out into a new file which just contains this class?

kano_updater/monitor.py
```diff
@@ -0,0 +1,165 @@
...
+
+    Returns true if we timed out and false if the process finished
+
+
+    """
+    watchpid = watchproc.pid
+
+    spoll = kano_updater.utils.signalPoll(signal.SIGUSR1)
+
+    lastEvent = time.time()
```
tombettany: `last_event`

kano_updater/monitor.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+def run(cmdargs):
+    os.environ["MONITOR_PID"] = str(os.getpid())
+    subproc = subprocess.Popen(cmdargs,
+                               shell=False)
+
+    if not monitor(subproc, MONITOR_TIMEOUT):
+        return subproc.returncode
+    else:
```
tombettany: Given that you are returning above, you don't need this `else` which will simplify the indentation of the code below.

kano_updater/monitor.py
Ealdwulf: That is indeed the wrong return code. Now fixed.

kano_updater/monitor.py
Ealdwulf: I've changed it to do that - in the gui case only, is that sufficient?

kano_updater/monitor.py
Ealdwulf: Woah, that's wong, I must have broken it when tidying the indentation.

kano_updater/monitor.py
Ealdwulf: Actually this function is performing a transitive closure. It's not a closure itself except in a trivial sense. But I'll rename it to transitive_closure.

kano_updater/monitor.py
skarbat: I think that's okay - the reason I used `_exit` was due to being in the middle of a Gtk event loop, I don't think that's the case here, so 👍

Ealdwulf:
@radujipa @tombettany I believe I've fixed all the above issues, except moving the test code @tombettany which I will do it a later commit.

bin/kano-updater-internal
```diff
@@ -0,0 +1,341 @@
+#!/usr/bin/env python
+
+# kano-updater
```
radujipa: *kano-updater-internal

tests/monitor_stub.py
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env python
+
```
radujipa: Copyright header?

kano_updater/signal_handling.py
```diff
@@ -0,0 +1,18 @@
+import signal
```
radujipa: Copyright header?

tests/monitor_stub.py
Ealdwulf: Fixed.

kano_updater/signal_handling.py
Ealdwulf: Fixed.

PR #299, opened 2018-03-13
Pinning down package versions

wizofe:
Changing the install target of pip-transition package to install
the required pth file. Creating a new pinning file so apt installer
accepts downgrades of packages as in current version.


wizofe:

conf/apt/pip
```diff
@@ -0,0 +1,36 @@
+Package: python-docopt
```
Ealdwulf: I think the following might be prefereable:
```
Package: python-*
Pin: origin dev.kano.me
Pin-Priority: 700

Package: python-*
Pin: origin repo.kano.me
Pin-Priority: 700
```
I think that would allow us to more easily remove the pinned versions without having to update this file.

wizofe: The `-*` seems intriguing. The problem though is that if `Pin-Priority` is under 990 then it will keep updating with the latest version. And if I assign everything higher than 1000 then it will keep downgrading all the `python-` packages.

Ealdwulf: Is that because you have them installed already?
I don't have them installed, and it seems to be fine trying to install the version of kano-updater from scratch.

wizofe: Very true, I have them and it should work on a clean image but again it's more secure to pin versions individually in case advanced users messed something up, I am thinking. If you feel it's preferable to go with this though, please comment.

Ealdwulf: You're right, if I change the above to 
```Package: python-*
Pin: origin dev.kano.me
Pin-Priority: 1000

Package: python-*
Pin: origin repo.kano.me
Pin-Priority: 1000
```

Then it installs python-docopt 0.6.2 even if 0.6.2-1 is already installed. I think this is prefereable to specifying the exact versions in the pin, if we want to adopt a version from raspbian we can stop exporting a version from our servers. Whereas if we specify versions then we always have to update this file.

wizofe: 👌

Ealdwulf: We both commented at the same time: regarding versions, we will get the right versions, because they are in the debian control file of kano-updater. I think it' better not to specify the versions in two different places, here and there, because if we have an inconsistency at any point it may refuse to install anything.

tombettany: I'm for Alex's suggestion of setting a prio for the source. The only problem might be what happens with [these priorities](https://github.com/KanoComputing/kano-updater/blob/master/conf/apt/preferences) which set suite priority - I don't know how the two will behave together.

wizofe: @tombettany That would be no problem as long as we have inside our repo the current `python-*` as 1000 takes precedence over the `preferences` priorities. OK, I'll bring the changes in.

PR #301, opened 2018-03-14
Fix package dependencies based on exact version numbers

skarbat:
Not exactly sure if that's the correct way to fix it, but it's due to this:

```
root@KXBN001:/# apt-get install kano-os-pip-transition
Reading package lists... Done
Building dependency tree
Reading state information... Done
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 kano-os-pip-transition : Depends: python-pam (= 0.1.4) but 0.4.2-13.1 is to be installed
E: Unable to correct problems, you have held broken packages.
```


radujipa:
#YOLO

wizofe:
Oh la la

PR #302, opened 2018-03-14
WIP Detach tty

Ealdwulf:
Before I forget, I think this could make the updater slightly more immune to packages asking interactive questions.
NOT proposing to merge for this rc, we've done enough to the updater already.

This PR:
 -  causes the updater to detach from its controlling terminal ( if any) tty to prevent any packages thinking they have interactive input.

### DIFF
--- a/kano_updater/utils.py
+++ b/kano_updater/utils.py
@@ -15,6 +15,8 @@ import grp
 import signal
 import re
 import subprocess
+import fcntl
+import termios
 
 from kano.logging import logger
 from kano.utils.shell import run_cmd, run_bg, run_cmd_log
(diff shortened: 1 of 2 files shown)


kano_updater/utils.py
```diff
@@ -577,3 +579,15 @@ def verify_kit_is_plugged():
...
 
     return is_plugged and not is_battery_
+
+
+def detach_from_tty():
+    """
+    Detach from controlling tty
+    """
+    try:
+        myTTY = os.open("/dev/tty",os.O_RDWR)
```
skarbat: Should this be broadly better if queried dynamically? To support docker interactive builds.

```
kanux@minos:~$ peldins-jessie
Started container: peldins-jessie with command: 'bash'
update-binfmts: warning: qemu-arm already enabled in kernel.
kanux@4f73998cc87b:~/peldins-jessie$ tty
/dev/pts/12
```

Ealdwulf: `/dev/tty` is a magic file which represents the tty attached to the current process.

skarbat: Enlightning... 👍

radujipa:
@Ealdwulf I've manually added your changes on top of the current `rc` branch since rebasing it was too much of a pain. I've kept a copy of the original branch at `detach_from_tty_old`.

PR #311, opened 2018-03-28
Use namespaced py2deb converted package of pam==0.1.4 from v3.14.1

radujipa:
The debian version of python-pam is wildly different to the version
we used to get from pypy. The former is a comprehensive python
wrapper for the C libpam and the latter is a simpler wrapper which
gives us a simple function for authentication and so its higher
level for authentication than its debian counterpart. While in the
future we may wish to move to python-pam from debian, for now it's
safer to revert to our previous version.


radujipa:

PR #312, opened 2018-03-29
Clean the apt cache after installation

radujipa:
It seems the apt cache is not cleaned completely with autoremove
after a download was completed. Upgrading an image from v2.4.0
shows some 500MB could be freed and autoclean is not helping with
that. Running the apt clean command after an update installation
should help with keeping the cache down reliably.


radujipa:
[(2) As a user, I want the updater to autoclean old cached files [2]](https://trello.com/c/2rgwr3GT/1744-2-as-a-user-i-want-the-updater-to-autoclean-old-cached-files-2)

PR #315, opened 2018-04-03
WIP raw clock

Ealdwulf:
This PR changes monitor.py to using raw clock to avoid premature timeout when the network comes up and sets the wall clock time.

Not tested yet.

### DIFF
--- a/kano_updater/raw_clock.py
+++ b/kano_updater/raw_clock.py
@@ -1,0 +1,36 @@
+# raw_clock.py
+#
+# Copyright (C) 2018 Kano Computing Ltd.
+# License: http://www.gnu.org/licenses/gpl-2.0.txt GNU GPL v2
+#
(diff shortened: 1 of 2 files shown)


kano_updater/raw_clock.py
```diff
@@ -0,0 +1,36 @@
...
+    _fields_ = [
+        ('tv_sec', ctypes.c_long),
+        ('tv_nsec', ctypes.c_long)
+    ]
+
+
+class raw_clock:
+    CLOCK_MONOTONIC_RAW = 4  # see <linux/time.h>
+    def __init__(self):
+        self.librt = ctypes.CDLL('librt.so.1', use_errno=True)
```
skarbat: Confirmed it's the same libname on debian stretch, could perhaps be useful to have it as a default parameter, as in `init(libname)` or so.

PR #316, opened 2018-04-05
Bump OS version to v4.0.0

radujipa:


radujipa:
[(1) As an engineer, I want to bump the OS version to 4.0.0 for the upcoming release](https://trello.com/c/pCNTZvqA/1839-1-as-an-engineer-i-want-to-bump-the-os-version-to-400-for-the-upcoming-release)

PR #325, opened 2018-07-04
Update deps for stretch and fix req disk space func

radujipa:
### Fix get_required_upgrade_space manual pkg marking and reverting

This issue presented itself when upgrading to Stretch, so it will
be required in an OS version before transitioning to Stretch. The
issue came about when manually marking and reverting pkg states
presumably in an order which caused a apt broken pkgs state. Use
the upgrade(dist_upgrade=True) and clear() functions of the
apt.cache to do the same safer and much much quicker.

### Bump pkg deps for Stretch

These changes are required on the Strech suite, not Jessie. This
updates package dependencies to avoid crashes and styling issues
in a Stretch Updater during the upgrade from Jessie.

---

Note: I've separated changes required for 3.16.0 and for 4.0.0 in separate commits.


radujipa:
[Ensure that in the future, As a RPi 2/3 (TBC) Jessie user, I will be able to see an update to Stretch](https://trello.com/c/4HwMCjH7/1898-ensure-that-in-the-future-as-a-rpi-2-3-tbc-jessie-user-i-will-be-able-to-see-an-update-to-stretch)

PR #326, opened 2018-07-04
Autoexpand: Fix autoexpand script

tombettany:
With the introduction of the recovery partition, more needs to be done
to support auto-expanding the SD card as there are additional partitions
to contend with while still maintaining backwards-compatibility with the
standard image layout. To do this, split the expand out into a set of
Python functions to simplify.


bin/expand-rootfs
```diff
@@ -1,79 +1,42 @@
...
+# Copyright (C) 2013 - 2018 Kano Computing Ltd.
+# License: http://www.gnu.org/licenses/gpl-2.0.txt GNU GPLv2
 #
 # Used in Kanux during interactive partition resizing - kano-update
-# rc=0 means resize scheduled for next reboot
-# rc=1 means partition already resized
-# any other value means error, look at syslog for details
+#
+# rc = 0 means resize scheduled for next reboot
+# rc = 1 means partition already resized
```
skarbat: can we add here "rc=10 means you need root permissions" ?

tombettany:
Thanks @skarbat. I've added more information about the return codes. To answer your other question, the tests will be run automatically when any change is pushed to Github. To others, I'm going to merge now so that it makes it into the image but still make any comments that you wish and we can change it.

skarbat:
👍

PR #327, opened 2018-07-06
Configure dpkg to run in noninteractive mode to avoid freezes

radujipa:
This tells dpkg and anything using it to run in noninteractive
mode such that it doesn't use programs like 'whiptail' to ask for
user input which is not supported by the Updater.

This fixes the 3.15.0 to 4.0.0 update getting stuck when configuring libc.


radujipa:
@tombettany This is the most reliable one I've found, check this out https://github.com/phusion/baseimage-docker/issues/58

radujipa:
[Ensure that in the future, As a RPi 2/3 (TBC) Jessie user, I will be able to see an update to Stretch](https://trello.com/c/4HwMCjH7/1898-ensure-that-in-the-future-as-a-rpi-2-3-tbc-jessie-user-i-will-be-able-to-see-an-update-to-stretch)

skarbat:
@tombettany , I  think @radujipa 's solution is a safer temporary method instead of a system-wide change...

```
To reconfigure debconf, use the dpkg-reconfigure tool from the debconf package; the relevant command is dpkg-reconfigure debconf. Note that the configured values can be temporarily overridden with environment variables when needed (for instance, DEBIAN_FRONTEND controls the interface, as documented in the debconf(7) manual page).
```

PR #330, opened 2018-08-14
 [OS-227] Stretch Update: Add upgrade path to v4.x.x

tombettany:
Up to now the Debian sources lists have been unmanaged and placed in the
system manually, migrate them into a package to allow them to be
properly managed, recoverable and verifiable.

Add a post-update step to the updater to switch the sources over to the
new repository for those users who have hardware compatible with v4.x.x
and relaunch the updater flow once those are in place to trigger the
update. For those staying on Jessie, 3.16.0 will be the final update and
the updater will clean up to ensure that all is well.

This will be in the `stretch` repository, a similar change will follow to be placed in `jessie`.


debian/kano-updater.postinst
```diff
@@ -35,16 +34,6 @@ case "$1" in
...
-        # Add the urgent repo to sources - if not already existent (jessie users)
-        grep -q 'release-urgent' $KANO_SOURCES || \
-            grep -m 1 '^deb http://.*.kano.me/archive-jessie/' $KANO_SOURCES | \
-                sed -r "s|^deb http://([a-z]*).kano.me.*$|deb http://\1.kano.me/archive-jessie/ release-urgent main|" >> $KANO_SOURCES
-
-        # Add the urgent repo to sources - if not already existent (wheezy users)
-        grep -q 'release-urgent' $KANO_SOURCES || \
-            grep -m 1 '^deb http://.*.kano.me/archive/' $KANO_SOURCES | \
-                sed -r "s|^deb http://([a-z]*).kano.me.*$|deb http://\1.kano.me/archive/ release-urgent main|" >> $KANO_SOURCES
-
```
radujipa: 👍 👍 👍

kano_updater/scenarios.py
```diff
@@ -797,7 +837,7 @@ def _bootconfig_set_value_helper(self, setting, value):
         except ImportError:
             logger.error("end_config_transaciton not present - update to kano-settings failed?")
 
-    def beta_370_to_beta_380(self):
+    def beta_370_to_beta_380(self, dummy_progress):
```
radujipa: I like that the scenarios are now able to report on progress back to the GUI. This scenario currently is a problem because it installs 11 apps without any progress indication. Could we create a ticket to fix this for later?

conf/apt/apt.conf.d/00-kano-default-release
```diff
@@ -0,0 +1 @@
+APT::Default-Release "release";
```
radujipa: Is this how we can switch between different sources now?

tombettany: Yeah, you need to do two things:

 - Create a new file in something like  `/etc/apt/apt.conf.d/99-release` (anything with a higher number than `00-*`) to say `APT::Default-Release "rc";`, for example
 - Create a new file in something like  `/etc/apt/preferences.d/99-pinning` (anything with a higher number than `00-*`) to unpin the other repo, for example:
 
```
Package: *
Pin: release n=rc
Pin-Priority: 500
```

These two steps are handled by `kano-dev`.

In addition, there is the really nice thing that you can directly install a package from `scratch` say without changing sources with:

```
apt-get install -t scratch my-package
```

conf/apt/sources.list.d/kano-jessie.list
```diff
@@ -0,0 +1,32 @@
+#
+# Kano OS Jessie Official Package Repository
+# Key:
+#   curl -L http://repo.kano.me/archive-stretch/repo.gpg.key | apt-key add -
+#
+
+deb http://repo.kano.me/archive-jessie/ release main
+deb http://dev.kano.me/archive-jessie/ devel main
```
radujipa: Let's create a test case where we make sure that when the dev server goes offline, the Updater doesn't report an error and exit.

PR #332, opened 2018-08-30
Revert pinning mechanism

Ealdwulf:
This PR reverts the use of APT:Default and pinning to switch to the development repos and suites.
It also introduces a mechanism to QA the full update from 3.xx to 3.16 to 4.1.0 (a path which the updater install insted of the production sources file for 4.1.0, so we can use one pointing at devel).

Not yet tested.
There will be two companion PRs, one for the jessie branch and one in peldins for kano-dev.


skarbat:
@Ealdwulf sorry, one moment... how is the `qa/kano-stretch-qa-test.list` going to be populated on the QA framework? I can't see how we are packaging it... and previous versions will not have it either

Ealdwulf:
@skarbat I hadn't considered that yet. However you want! We can move it to the osupgrades repo if you prefer.

skarbat:
It's a similar dilemma we have with `kano-dev`, I'll move a copy to the osupgrades as well, leave it to me @Ealdwulf , PR approved for me, and we can go ahead - Thanks! :)

wizofe:
Last time @Ealdwulf you manually copied the `list` file through your script, right? I think it would be great to move a copy to `osupgrades` repo, so we have it available for the tests there!

Ealdwulf:
Yes, right now there is a copy on the osupgrades server which I just put
tehre manually.

Ealdwulf:
Okay, I reverted pushing this file to kano-updater, so it's only in one
place now.

Ealdwulf:
Given that out of 3 updates 2 seemed to work, and the remaining one didn't seem to be caused by the update path, I am going to merge these to reduce the amount of changes in flight.

PR #334, opened 2018-08-30
[OS-273] Ensure that the sources list is re-read and updated correctly.

wizofe:
The sources file list is renewed using the Cache class of the apt library but it seems that this is not working always as intended. It's possible that we are missing the re-read state of the packages. This is an attempt to fix it here and if that is not successful then it will be added during the download phase of the updater as well.


wizofe:
:gun: due to fixes from #332 #333

PR #335, opened 2018-09-07
Remove vestigial symlink.

Ealdwulf:
Remove dangling symlink left by #333 .


skarbat:
I'll keep an eye on this one, because...

```dr info debuild gcc -Wall `pkg-config --cflags lxpanel gtk+-2.0` -shared -fPIC kano_updater.c parson/parson.c -o kano_updater.so `pkg-config --libs lxpanel gtk+-2.0` -lkdesk-hourglass

dr info debuild Package lxpanel was not found in the pkg-config search path.

dr info debuild Perhaps you should add the directory containing `lxpanel.pc'
```

Ealdwulf:
FYI, this is now pushed to the devel suite of jessie in `dr`.

Ealdwulf:
@skarbat That error's because it always tries to build it for stretch now - it can't tell from the branch name that this is intended for jessie :-). We have to manually trigger build and push of jessie branches, as I did above.

skarbat:
Oh I see  - case is cleanly resolved now, thanks @Ealdwulf !

PR #337, opened 2018-09-13
Jessie update fixes nodebug

Ealdwulf:
This PR is the jessie version of the changes I have been working on.
For ease of review it does not contain all the debug prints, which I have moved to a separate PR. 
The version we have tested combines both.
The changes are:
 - launch the splash without adding a monitor
 - set the app name of the splash so it writes to a separate log file
 - make the install* functions call relaunch and return false, instead of raising the exception there.
 - make the 4.1 scenario raise an exception instead of calling progress.relaunch(), since it will be ignored if it is not in the right thread.
 - make the scenario call apt-get update as a script instead of inside python (not sure this was actually necessary, but we tested it)


kano_updater/utils.py
```diff
@@ -237,12 +237,12 @@ def _handle_sigusr1(signum, frame):
 
 
 def show_relaunch_splash():
-    cmd = ["kano-updater", "ui", "relaunch-splash", str(os.getpid())]
-    p = subprocess.Popen(cmd, shell=False)
-
     # register a handler for SIGUSR1
     signal.signal(signal.SIGUSR1, _handle_sigusr1)
 
+    cmd = ["kano-updater-internal", "ui", "relaunch-splash", str(os.getpid())]
```
skarbat: Pretty sure you've checked that - but are the double dashes for these options getting magically in?

Ealdwulf: These options don't take double dashes :-)

PR #349, opened 2018-12-04
Manually reverted v4.1.2 version bump

radujipa:
As v4.1.2 will not go out anymore, let's just remove it.


radujipa:
@tombettany Removed the test, porting the 3.16.1 changes from Jessie will happen on a new PR

PR #355, opened 2018-12-05
[OS-578] - Resolve broken wireless kernel driver by updating firmware package

skarbat:
This PR fixes a problem whereby upgrading from a 3.6.0 (and possibly any 3.x version) would leave the wireless driver software broken, the wireless device undetectable, and thereby breaking wireless networking for usual work, and specially troublesome during an updater recovery session.

Not entirely sure this is the best callback hook to install the latest firmware, but this should fix the problem in the 3.6.0 upgrade path that I have tested.

This PR should go against Jessie, and the package pushed to that repo as well - not Stretch

cc @wizofe @radujipa @tombettany


Ealdwulf:
Is there a reason not to make this a proper dependency? This way it won't be prefetched in the background.

skarbat:
Thanks for comments @Ealdwulf - when you say a proper dependency, do you mean to say we should make a specific kano OS package depend on this one? Like `kano-settings` for instance?

radujipa:
I agree with @Ealdwulf, this isn't reliable enough. Don't remember if we have [kano-platform-essential](https://github.com/KanoComputing/kano-platform-essential/blob/master/debian/control) in jessie, but there should be a deb dependency on that firmware.

Ealdwulf:
Yes @skarbat, that is what I meant.

skarbat:
The problem I see with using `kano-platform-essential` package is two-fold:

 * It's not available on Jessie, specifically 3.16.0
 * The fix comes by installing `firmware-brcm2011 v1:20161130` from our stretch mirror

Would it be safer to upgrade this package right after we introduce the Stretch APT sources?

tombettany:
A few things: first we might just want to backport the `kano-platform-essential` package (essentially push it as-is, or with a version change, to a `jessie` branch, build the branch to the Jessie repo).

Second, I don't think that it is a good idea to automatically install the new firmware package after we introduce the Stretch sources for a few reasons:
 - We'll be splitting the upgrade from 3.x -> 4.x into two, the 3.x should be only Jessie, nothing else (a half-and-half system will be impossible to guarantee reliability and debug)
 - The final version of Jessie provided by RPi Foundation *definitely* works with the boards that we are shipping (as they wouldn't release without it working) so there is no reason why we can't get it to work within the confines of Jessie.

Is the problem that package goes missing or that it was never installed in the first place?

skarbat:
The situation at the time the wireless device is lost is as follows

 * The system is at Debian Jessie - kano os v3.16.0
 * The firmware package is installed `firmware-brcm80211`
 * For some mystery, it's not providing the file `/lib/firmware/brcm/brcmfmac43430-sdio.txt`
 * The kernel module needs this `txt` file in order to bring up wlan0.

So it would appear we have two packages out of sync for some reason, but definitely upgrading the firmware from the Stretch repo fixes the problem.

If we introduce this package as a dependency, we'd have to specify the correct version, and that makes me think we can easily break the apt installation flow because it's only visible under Stretch.

tombettany:
I'm closing this because this "solves" the problem by pulling from Stretch - we should resolve our Jessie repository so this is resolved entirely via `apt` and if we do so we don't even need to backport the `kano-platform-essential` as the problem is just that the wrong version of the package is getting installed.

Further discussion to happen on the ticket itself: [OS-578](https://kanocomputing.atlassian.net/browse/OS-578).

PR #357, opened 2018-12-05
[OS-604] [Stretch] Sources: Move sources file to kano-repos.list

tombettany:
The Stretch sources file was originally stored as
`/etc/apt/sources.list.d/kano-stretch.list` but this is very specific to
the base system so make it more general. Also fix the link between this
file and the trigger for checking if updates are available.

Probably should also backport this to 3.16.1.


tests/test_sources.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+import os
+import re
+
+
+def test_paths_matches_install(apt_sources_install):
+    from kano_updater.paths import KANO_SOURCES_LIST
+
+    assert os.path.exists(KANO_SOURCES_LIST)
```
radujipa: - Let's move REFERENCE_STRETCH_LIST from scenarios.py (as well as other appropriate ones) into paths.py
 - Let's add another assert for REFERENCE_STRETCH_LIST file such that we know the symlink isn't broken

tombettany: Moved the sources dir, `OS_SOURCES_REFERENCE` and `REFERENCE_STRETCH_LIST` to `paths.py` and added a test assertion that the references exist.

PR #361, opened 2018-12-09
[OS-571, OS-616] [Stretch] Logging: Increase the logging level

tombettany:
Make crash reports more useful by providing more detail about where
things went wrong.

Note, everything at or above logging level `info` is logged to the filesystem in the updater.

Backported in https://github.com/KanoComputing/kano-updater/pull/362


tombettany:
Feedback is definitely still welcome but I'm going to merge so we can test tomorrow.

PR #362, opened 2018-12-09
[OS-571, OS-616] [Jessie] Logging: Increase the logging level

tombettany:
Backport to Jessie of https://github.com/KanoComputing/kano-updater/pull/361.

Make crash reports more useful by providing more detail about where
things went wrong.

Note, everything at or above logging level info is logged to the filesystem in the updater.


tombettany:
Merging with #361

PR #363, opened 2018-12-09
[OS-618, OS-571] Reporting: Send crash reports on fail

tombettany:
Attempt to assemble and send crash reports to us for analysis any time
there is an update aborted or error created.

We might want to consider backporting this but I'll leave that decision for tomorrow.


tombettany:
Likewise, merging in for testing.

PR #366, opened 2018-12-11
[OS-620] [Stretch] Apt: Retry downloads before failing

tombettany:
Retry downloads after download errors before failing completely.

I raise wherever the original would have raised - don't know if that's a good idea or not so feedback here is welcome. Only tested in as far as the tests cover, not done a proper update test with it.


kano_updater/apt_wrapper.py
```diff
@@ -99,6 +100,51 @@ def refresh_instance(self):
...
 
+    def _update_cache(self, progress, src_count, sources_list):
+        try:
+            self._do_update_cache(progress, src_count, sources_list)
+        except (
+                apt.cache.FetchFailedException, AptDownloadFailException
+        ) as err:
+            # Continue regardless, if it fails later because it can't grab a
+            # package then let it fail because of that; now that we've added
+            # in the development repositories it is more likely to happen.
```
radujipa: added in the dev repos?? What do you mean?

kano_updater/retry.py
```diff
@@ -0,0 +1,56 @@
...
+    def deco_retry(f):
+
+        @wraps(f)
+        def f_retry(*args, **kwargs):
+            mtries, mdelay = tries, delay
+            while mtries > 1:
+                try:
+                    return f(*args, **kwargs)
+                except ExceptionToCheck, e:
+                    msg = "%s, Retrying in %d seconds..." % (str(e), mdelay)
```
radujipa: could use .format() instead

kano_updater/retry.py
```diff
@@ -0,0 +1,56 @@
...
+                    logger.warn(msg)
+                    time.sleep(mdelay)
```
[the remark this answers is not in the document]
tombettany: I'd pulled that retry guy as-is from a Python article - relying on their QA :p but yeah, definitely a good point here - I'll add a heartbeat.

Also, don't have a problem with increasing the retry count.

kano_updater/retry.py
tombettany: Yeah, I forgot to fix that as I copied it :p

kano_updater/apt_wrapper.py
tombettany: I just moved the comment from before, I'll fix it up with an accurate comment.

tests/test_apt_wrapper.py
```diff
@@ -59,3 +59,88 @@ def test_space(apt):
...
+    if raise_expected:
+        with pytest.raises(AptDownloadFailException):
+            wrapper._fetch_archives(progress)
+    else:
+        wrapper._fetch_archives(progress)
+
+    assert fetch_archives.call_count == min(EXPECTED_FAILS + 1, MAX_RETRIES)
+
+
+def test_updating_cache(apt, mocker, monkeypatch):
```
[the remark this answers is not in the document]
tombettany: On it.

tombettany:
I'll fix up and backport. Didn't want to have to backport multiple times.

PR #370, opened 2019-01-02
[OS-827] Debian: Add apt HTTPS dependency

tombettany:
Add dependency on the `apt-transport-https` package to allow downloading
from secure sources should they need to be.

Also bump the updater version to 4.2.1.


debian/control
```diff
@@ -30,6 +30,7 @@ Depends:
     libkdesk-dev,
     gir1.2-gtk-3.0 (>= 3.22.11),
     kano-settings (>= 1.3-1),
+    apt-transport-https,
```
radujipa: Let's add the min version of the package here `(>= 1.4.8)` so we know in the future what it used to be when it changes. I think it's good practice to always specify a min version.

tests/fixtures/version.py
```diff
@@ -64,6 +64,7 @@
     OSVersion(devstage='Beta', version='4.1.0', name='Hopper'),
     OSVersion(devstage='Beta', version='4.1.1', name='Hopper'),
     OSVersion(devstage='Beta', version='4.2.0', name='Hopper'),
+    OSVersion(devstage='Beta', version='4.2.1', name='Hopper'),
```
radujipa: Could bump the copyright year here too

tombettany: Done

debian/control
tombettany: Good idea.

PR #372, opened 2019-01-07
[OS-839] Fixed a few tbks when logging Exceptions

radujipa:
The issue seems to be caused by sending an Exception object to the logger instead of its string representation.

I've gone through all calls to the `logger` for this to make sure there weren't any other ones lurking in the shadows.


kano_updater/apt_wrapper.py
```diff
@@ -355,8 +355,8 @@ def fix_broken(self, progress):
         if self._cache._depcache.broken_count:
             try:
                 self._cache._depcache.fix_broken()
-            except SystemError as err:
-                logger.error(err)
+            except SystemError as e:
+                logger.error('Error attempting to fix broken pkgs {}'.format(e))
```
Ealdwulf: Maybe it would be more useful to do:
```suggestion
                logger.error('Error attempting to fix broken pkgs', exception=e)
```
And similarly below.

radujipa: Good point, Alex. I'll use that instead.

PR #373, opened 2019-01-10
[OS-861] Logged a message when the Updater receives a timeout from the monitor

radujipa:
It's pretty hard to tell when looking at the logs when the timeout occurred.


tombettany:
Can we create/link a JIRA ticket so we can track testing this?

radujipa:
@tombettany I've made sure we flush the logs before sending crash logs here. It's been ported to Jessie as well here https://github.com/KanoComputing/kano-updater/pull/376

PR #374, opened 2019-01-10
Fix GPG key URL comments in sources files

radujipa:
Small typos here and there


radujipa:
It's just comments, relax 😄

skarbat:
:-)

PR #379, opened 2019-01-17
[OS-839] [Backport] Fixed a few tbks when logging Exceptions

radujipa:
The issue seems to be caused by sending an Exception object
to the logger instead of its string representation.

Backport of https://github.com/KanoComputing/kano-updater/pull/372


kano_updater/apt_wrapper.py
```diff
@@ -355,8 +355,8 @@ def fix_broken(self, progress):
         if self._cache._depcache.broken_count:
             try:
                 self._cache._depcache.fix_broken()
-            except SystemError as err:
-                logger.error(err)
+            except SystemError as e:
+                logger.error('Error attempting to fix broken pkgs', exception=e)
```
skarbat: I've seen us doing it like this at several places: `logger.error('ouch: {}'.format(e))`, are both methods valid?

skarbat: 👍

PR #382, opened 2019-01-17
[OS-859] [Backport] Fixed kano-profile dependency for import touch() tbk

radujipa:
Previously there was a fix in Profile on 3.16.0 but the dependency
in the Updater was never updated.

Was fixed here https://github.com/KanoComputing/kano-profile/pull/464

Backport for Jessie of https://github.com/KanoComputing/kano-updater/pull/381


skarbat:
Some tests failing though, and compilation too it seems...

```kano_updater.c:9:21: fatal error: gtk/gtk.h: No such file or directory```

PR #384, opened 2019-01-17
[OS-857] Added 10s cooldown to launching Flappy to prevent many instances

radujipa:
This should help prevent putting extra unnecessary stress on
system resources.


kano_updater/ui/views/install.py
```diff
@@ -119,9 +122,13 @@ def _create_warning_label(self):
 
     def _launch_game(self, window=None, event=None):
         """ Key release event handler for this view which only launches the game """
+        timestamp = time.time()
         try:
-            if event and event.get_keyval()[1] in [Gdk.KEY_j, Gdk.KEY_J]:
-                os.system('{} &'.format(FLAPPY_PATH))
+            if event and event.get_keyval()[1] in [Gdk.KEY_j, Gdk.KEY_J] and \
+               (self.game_launch_timestamp + self.FLAPPY_LAUNCH_COOLDOWN) < timestamp:
```
skarbat: Are we sure this is okay? Unix time is number of elapsed seconds since January 1st 1970 if I'm not mistaken.

radujipa: It'd probably be more readable if I reverse the statement: `if t(t+1) > t(t) + 10s then launch_game()`.
I'll update both Jessie and Stretch.

skarbat: 👍

PR #389, opened 2019-01-21
[OS-860] Added free disk space check before the download phase

radujipa:
This will make sure that the Updater doesn't download in the background to the point where the disk becomes completely full.

NB: Coupled with https://github.com/KanoComputing/kano-updater/pull/387 it may cause people to abort an update after the Updater updates itself.


kano_updater/commands/download.py
```diff
@@ -76,21 +78,34 @@ def download(progress=None, gui=True):
...
+
     if not is_server_available():
         err_msg = N_("Could not connect to the download server")
         logger.error(err_msg)
         progress.fail(_(err_msg))
         RCState.get_instance().rc = RC.CANNOT_REACH_KANO
         return False
 
+    run_cmd('sudo kano-empty-trash')
+    enough_space, space_msg = check_disk_space(priority)
```
tombettany: Not necessarily a problem but the`install()` function calls the `download()` function and expects a `True` return to continue with its things - this means that the `install()` will end up checking for disk space twice within a few seconds of one another.

radujipa: That's true. Let's do the check twice during an install for now and see how that performs. Can't hurt tbh.

PR #394, opened 2019-01-30
[OS-268] [Jessie] PreUpdate workaround for installing firmware-ralink over firmware-misc-nonfree

radujipa:
Because we did not have in place mirrors for Raspbian and our RaspberryPi mirrors were old, users in Jessie started installing `deb8u` pkgs over `rpi` ones. After updating our RPi mirrors to align with Raspbian ones, there was an issue updating from 3.16.1 to 3.16.2.

```
Error Installing packages: /var/cache/apt/archives/firmware-ralink_1%3a0.43+rpi6_all.deb: trying to overwrite '/lib/firmware/rt2661.bin', which is also in package firmware-misc-nonfree 20161130-4~deb8u1
```

Essentially, there seems to be a file ownership transfer from `firmware-misc-nonfree` (deb8) to the `firmware-ralink` (rpi) package. May be worth raising this upstream. For now, attempting workaround by removing the package in a PreUpdate scenario.


radujipa:
Merging this PR temporarily in order to test the first once more. I don't believe my previous build was sound given that the system collapsed without there being an update issue.

PR #398, opened 2019-02-12
[OS-844] Removed v3.7 to v3.8 3P app installation disk space requirement

radujipa:
Before, the disk space requirement was part of the update itself.
Moved away from that model to install apps separately of the update
to minimise the requirement (divide and conquer).
However, the App Store does not check for disk space before
installing an app so account for this here naively.
Requirements (disk_req) in MB were made with apt on a clean system.

### DIFF
--- a/kano_updater/ui/media/flappy-judoka/src/utils.py
+++ b/kano_updater/ui/media/flappy-judoka/src/utils.py
@@ -1,9 +1,7 @@
-
 # utils.py
 #
-# Copyright (C) 2015-2016 Kano Computing Ltd.
+# Copyright (C) 2015-2019 Kano Computing Ltd.
 # License: http://www.gnu.org/licenses/gpl-2.0.txt GNU GPL v2
-#
 #
 # Various utilities used throught the game.
 
@@ -78,7 +76,7 @@ def load_image(name, colorkey=None, alpha=False):
     else:
         image = image.convert()
     if colorkey is not None:
-        if colorkey is -1:
+        if colorkey == -1:
             colorkey = image.get_at((0, 0))
         image.set_colorkey(colorkey, pygame.RLEACCEL)
     return image


kano_updater/scenarios.py
```diff
@@ -908,22 +908,53 @@ def beta_370_to_beta_380(self, dummy_progress):
...
-        for app in new_apps:
-            run_cmd_log('kano-apps install --no-gui {app}'.format(app=app))
+        try:
+            # Install 3rd party apps from Kano World using the App Store.
+            # Before, the disk space requirement was part of the update itself.
+            # Moved away from that model to install apps separately of the update
+            # to minimise the requirement (divide and conquer).
+            # However, the App Store does not check for disk space before
+            # installing an app so account for this here naively.
+            # Requirements (disk_req) in MB were made with apt on a clean system.
```
tombettany: Can we have a TODO to move this logic into `kano-apps`?

kano_updater/scenarios.py
```diff
@@ -908,22 +908,53 @@ def beta_370_to_beta_380(self, dummy_progress):
...
+                {'kw_app': 'libreoffice', 'disk_req': 385},
+            ]
+
+            run_cmd_log('apt-get autoremove -y')
+
+            for app in new_apps:
+                run_cmd_log('apt-get clean')
+
+                mb_free = get_free_space()
+                mb_required = app['disk_req'] + 200  # MB buffer
```
tombettany: Are we happy with 200MB or do we want to use the buffer from the `disk_requirements` module?

kano_updater/scenarios.py
```diff
@@ -908,22 +908,53 @@ def beta_370_to_beta_380(self, dummy_progress):
...
+                        "Cannot install {app} as it requires {mb_required} but"
+                        " only has {mb_free}"
+                        .format(
+                            app=app['kw_app'],
+                            mb_required=mb_required,
+                            mb_free=mb_free
+                        )
+                    )
+        except Exception:
+            logger.error("Failed to install 3rd party apps")
```
tombettany: Do we want to actually log this exception? Do we want to trigger a feedback report in this case so we know about it?

tombettany: Also, should we catch each app separately?

PR #409, opened 2019-09-02
[OS-1006] - Remove non COPPA compliant apps through the upgrade

skarbat:
This PR removes the non-COPPA compliant apps when upgrading Kano OS to 4.3.2 version.
Tested the steps on an upgraded 4.3.2 image manually, should work fine.


kano_updater/scenarios.py
```diff
@@ -1256,5 +1256,17 @@ def beta_4_3_0_to_beta_4_3_1(self, dummy_progress):
         pass
 
     def beta_4_3_1_to_beta_4_3_2(self, dummy_progress):
+        # Remove non COPPA compliant apps
+
+        # web whatsapp has no debian candidate, remove manually
+        run_cmd_log('rm -f /usr/share/applications/*whatsapp*')
+        run_cmd_log('/usr/share/icons/Kano/66x66/apps/whatsapp.png')
```
tombettany: Methinks that you are missing the actual command here - you're just running the file!

tombettany: Also, is this file just added through the app store or is it owned by a debian package?

skarbat: Ouch! Got catch! fixed missing command.
The file is not owned by any debian package, that's the thing.

PR #413, opened 2019-09-13
Re-enable Ultimate parental controls level after kano-settings fix

radujipa:
Requires: https://github.com/KanoComputing/kano-settings/pull/535


kano_updater/scenarios.py
```diff
@@ -1279,10 +1279,6 @@ def beta_4_3_1_to_beta_4_3_2(self, dummy_progress):
...
         # Set Parental Controls to Ultimate for all existing users. COPPA.
-        # FIXME: The parental command sets the parental level and then
-        #        restarts the sentry server, blocking on the process. For now
-        #        just disable this and we will re-enable later.
-        # run_for_every_user(
-        #     'sudo kano-settings-cli set parental --level=3 "kano"'
-        # )
-        pass
+        run_for_every_user(
+            'sudo kano-settings-cli set parental --level=3 "kano"'
```
skarbat: I am just realizing... there's no way we can possibly run this part of `kano-updater` without sudo permissions anyway.

skarbat (approved):
Aside from comment, looks good

kano_updater/scenarios.py
radujipa: That's true, but at a minimum it's explicit about the fact that the command needs to be run with elevated privileges.

